PR #4, opened 2018-12-17
dma channel cluster support

burrbull:


therealprof:
I've added a do-not-merge flag so people won't accidentally merge it.

PR #5, opened 2019-01-05
Add support for setting initial pin state

WGH-:
Allows to set pin to high or low before actual switch to the output mode, eliminating small "low" window in cases where pin should be idle high.

I have tested this with push-pull output. The window is apparent when viewed in a logic analyzer, and this change eliminates it.

Please review thoroughly, as I don't have much expertise in neither Rust nor STM32 :)

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -46,6 +46,11 @@ pub struct Alternate<MODE> {
     _mode: PhantomData<MODE>,
 }
 
+pub enum State {
+    High,
+    Low,
+}
+
 macro_rules! gpio {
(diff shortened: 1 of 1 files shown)


therealprof:
Thanks for the PR. That's a very nice idea! Implementationwise I'd prefer if the function could be implemented in a non-breaking, either by allowing the configuration on an unconfigured pin or by introducing a new method.

TeXitoi:
I think `into_alternate_push_pull_with_state` and friend will do the trick, with `into_alternate_push_pull` implemented by using `into_alternate_push_pull_with_state` with the normal default value.

Issue #8, opened 2019-01-07
The documentation link on crates.io doesn't work


Mendelt:
It looks like the documentation link in Cargo.toml is broken. It is "https://docs.rs/crate/stm32f1xx-hal" but I think it should be "https://docs.rs/stm32f1xx-hal" without crate.
Or you can just leave the documentation key out. Crates.io auto-generates a link to docs.rs if none is specified.
I'm not sure if this is the only problem though because the documentation doesn'tseem to have been generated on docs.rs at all for this crate.

TheZoq2:
Looks like the link was fixed in 67aa99e but the docs generation is still broken, most likely because the required changes to Cargo.toml (`[package.metadata.docs.rs]`) have not been published to docs.rs yet. @therealprof should we push a new version or is there something else that needs fixing before we do that?

Mendelt:
No I think you're right. I already saw the fix but was waiting for the change to be published. I think docs.rs pulls from crates.io so the docs should show up after publishing.

therealprof:
@TheZoq2 Yeah, we should also bump the `stm32f1` version.

PR #15, opened 2019-02-10
Update to stm32f1 0.6.0 and bump version to 0.2.0

TheZoq2:
As discussed in #8 

@therealprof How does publishing this to crates.io work, do I have the permissions to do so?

How does cargo package work when feature flags are required to build the crate? I tried adding --features as you would when building but that did not work.

Also, I assume that 0.2.0 is correct since our exported dependency had a minor version bump, is that correct?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ features = ["stm32f103", "rt"]
 cortex-m = "0.5.8"
 nb = "0.1.1"
 cortex-m-rt = "0.6.7"
-stm32f1 = "0.5.0"
+stm32f1 = "0.6.0"
 
 [dependencies.void]
 default-features = false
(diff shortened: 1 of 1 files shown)


therealprof:
Too late. ;)

therealprof:
> @therealprof How does publishing this to crates.io work, do I have the permissions to do so?

Nope, I'll have to add you as owner first to allow you to publish this crate. I can do that if you want, but you'll need to have an account with crates.io.

> How does cargo package work when feature flags are required to build the crate? I tried adding --features as you would when building but that did not work.

That does not work, yet. I think the next released version of cargo will allow to do that. For now you'd have to use `--no-verify`.

> Also, I assume that 0.2.0 is correct since our exported dependency had a minor version bump, is that correct?

That is correct.

TheZoq2:
> Too late. ;)

Haha, oops :P

>Nope, I'll have to add you as owner first to allow you to publish this crate. I can do that if you want, but you'll need to have an account with crates.io.

Sure, that might come in handy. I do have a crates.io account which is the same as my github I believe

PR #17, opened 2019-02-13
Add support for real time clock

TheZoq2:
This is essentially the same code as the one in https://github.com/japaric/stm32f103xx-hal/pull/98 and it still has some issues.

# Where to initialise
The main problem is that LSE initialization and rtc clock selection has to happen before clock freeze. However, in order to do that, the backup domain has to be enabled in the pwr register. Since freeze, or any other function dealing with rcc does not modify pwr, adding that as a parameter would be a breaking change.

Additionally, we want to statically ensure (I think) that the required clocks are on when the real time clock is started. There are a couple of solutions to this

## Token type
My current solution to all this is to add a function to rcc that sets up things in the backup domain which gives a token that is required in order to run the rtc.

I'm not a huge fan of that solution, but I can't think of anything that is obviously better.

## Rtc "owned" by `RCC`
One option would be to, rather generate a token, generate an `UninitialisedRtc` struct which can then be used to actually initialise the `RTC` by passing the `Clock` struct.

## Dynamic initialisation check

We could add a function to cfgr which enables the backup domain and adds something to clocks to indicate that that has been done. Then, the `RTC` would check wether this value is set. I'd say this is a bad solution, we should avoid runtime errors wherever possible.

# Incomplete implementation
The current implementation does not support changing the prescaler or clock speed which I think can be added later.

Does anyone have any thoughts on this? which solution do you prefer for initialisation?

### DIFF
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -30,7 +30,7 @@ fn main() -> ! {
     let backup_domain = rcc.cfgr.enable_backup_domain(&mut pwr);
     let clocks = rcc.cfgr.freeze(&mut flash.acr);
 
-    let rtc = backup_domain.rtc(p.RTC, clocks);
+    let rtc = backup_domain.rtc(p.RTC, &clocks);
 
     loop {
         writeln!(hstdout, "time: {}", rtc.read()).unwrap();
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -27,16 +27,14 @@ fn main() -> ! {
     let p = hal::stm32::Peripherals::take().unwrap();
 
     let mut pwr = p.PWR;
-    let mut flash = p.FLASH.constrain();
     let mut rcc = p.RCC.constrain();
     let backup_domain = rcc.bkp.constrain(p.BKP, &mut rcc.apb1, &mut pwr);
-    let _clocks = rcc.cfgr.freeze(&mut flash.acr);
     let lse = rcc.lse.freeze(&backup_domain);
 
     let rtc = Rtc::rtc(p.RTC, lse, &backup_domain);
 
     loop {
-        writeln!(hstdout, "time: {}", rtc.read_counts()).unwrap();
+        writeln!(hstdout, "time: {}", rtc.read_seconds()).unwrap();
     }
 }
 
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -29,7 +29,7 @@ fn main() -> ! {
     let mut pwr = p.PWR;
     let mut rcc = p.RCC.constrain();
     let mut backup_domain = rcc.bkp.constrain(p.BKP, &mut rcc.apb1, &mut pwr);
-    let lse = rcc.lse.freeze(&mut backup_domain);
+    let lse = backup_domain.enable_lse(rcc.lse);
 
     let rtc = Rtc::rtc(p.RTC, lse, &mut backup_domain);
 
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -27,7 +27,7 @@ fn main() -> ! {
     let mut pwr = p.PWR;
     let mut flash = p.FLASH.constrain();
     let mut rcc = p.RCC.constrain();
-    let backup_domain = rcc.cfgr.enable_backup_domain(&mut pwr);
+    let backup_domain = rcc.cfgr.access_backup_domain(&mut pwr);
     let clocks = rcc.cfgr.freeze(&mut flash.acr);
 
     let rtc = backup_domain.rtc(p.RTC, &clocks);
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -28,13 +28,13 @@ fn main() -> ! {
 
     let mut pwr = p.PWR;
     let mut rcc = p.RCC.constrain();
-    let backup_domain = rcc.bkp.constrain(p.BKP, &mut rcc.apb1, &mut pwr);
-    let lse = rcc.lse.freeze(&backup_domain);
+    let mut backup_domain = rcc.bkp.constrain(p.BKP, &mut rcc.apb1, &mut pwr);
+    let lse = rcc.lse.freeze(&mut backup_domain);
 
-    let rtc = Rtc::rtc(p.RTC, lse, &backup_domain);
+    let rtc = Rtc::rtc(p.RTC, lse, &mut backup_domain);
 
     loop {
-        writeln!(hstdout, "time: {}", rtc.read_seconds()).unwrap();
+        writeln!(hstdout, "time: {}", rtc.seconds()).unwrap();
     }
 }
 
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -18,6 +18,8 @@ use sh::hio;
 
 use core::fmt::Write;
 
+use hal::rtc::Rtc;
+
 #[entry]
 fn main() -> ! {
     let mut hstdout = hio::hstdout().unwrap();
@@ -27,10 +29,11 @@ fn main() -> ! {
     let mut pwr = p.PWR;
     let mut flash = p.FLASH.constrain();
     let mut rcc = p.RCC.constrain();
-    let backup_domain = rcc.backup_domain(&mut pwr);
-    let clocks = rcc.cfgr.freeze(&mut flash.acr);
+    let backup_domain = rcc.backup_domain(p.BKP, &mut pwr);
+    let _clocks = rcc.cfgr.freeze(&mut flash.acr);
+    let lse = rcc.lse.freeze(&backup_domain);
 
-    let rtc = backup_domain.rtc(p.RTC, &clocks);
+    let rtc = Rtc::rtc(p.RTC, lse, &backup_domain);
 
     loop {
         writeln!(hstdout, "time: {}", rtc.read_counts()).unwrap();
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -29,7 +29,7 @@ fn main() -> ! {
     let mut pwr = p.PWR;
     let mut flash = p.FLASH.constrain();
     let mut rcc = p.RCC.constrain();
-    let backup_domain = rcc.backup_domain(p.BKP, &mut pwr);
+    let backup_domain = rcc.bkp.constrain(p.BKP, &mut rcc.apb1, &mut pwr);
     let _clocks = rcc.cfgr.freeze(&mut flash.acr);
     let lse = rcc.lse.freeze(&backup_domain);
 
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -12,7 +12,6 @@ extern crate panic_semihosting;
 extern crate stm32f1xx_hal as hal;
 
 use hal::prelude::*;
-use hal::rtc::Rtc;
 use rt::{entry, exception, ExceptionFrame};
 
 use sh::hio;
@@ -23,16 +22,15 @@ use core::fmt::Write;
 fn main() -> ! {
     let mut hstdout = hio::hstdout().unwrap();
 
-    let mut p = hal::stm32::Peripherals::take().unwrap();
+    let p = hal::stm32::Peripherals::take().unwrap();
 
     let mut pwr = p.PWR;
     let mut flash = p.FLASH.constrain();
-    // Enable the clocks in the backup domain
-    let bd_token = p.RCC.enable_backup_domain(&mut pwr);
-    let rcc = p.RCC.constrain();
+    let mut rcc = p.RCC.constrain();
+    let backup_domain = rcc.cfgr.enable_backup_domain(&mut pwr);
     let clocks = rcc.cfgr.freeze(&mut flash.acr);
 
-    let rtc = Rtc::rtc(p.RTC, &bd_token, clocks);
+    let rtc = backup_domain.rtc(p.RTC, clocks);
 
     loop {
         writeln!(hstdout, "time: {}", rtc.read()).unwrap();
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -1,0 +1,50 @@
+//! Blinks an LED
+
+#![deny(unsafe_code)]
+#![deny(warnings)]
+#![no_std]
+#![no_main]
+
+extern crate cortex_m;
+extern crate cortex_m_rt as rt;
+extern crate cortex_m_semihosting as sh;
+extern crate panic_semihosting;
+extern crate stm32f1xx_hal as hal;
+
+use hal::prelude::*;
+use hal::rtc::Rtc;
+use rt::{entry, exception, ExceptionFrame};
+
+use sh::hio;
+
+use core::fmt::Write;
+
+#[entry]
+fn main() -> ! {
+    let mut hstdout = hio::hstdout().unwrap();
+
+    let mut p = hal::stm32::Peripherals::take().unwrap();
+
+    let mut pwr = p.PWR;
+    let mut flash = p.FLASH.constrain();
+    // Enable the clocks in the backup domain
+    let bd_token = p.RCC.enable_backup_domain(&mut pwr);
+    let rcc = p.RCC.constrain();
+    let clocks = rcc.cfgr.freeze(&mut flash.acr);
+
+    let rtc = Rtc::rtc(p.RTC, &bd_token, clocks);
+
+    loop {
+        writeln!(hstdout, "time: {}", rtc.read()).unwrap();
+    }
+}
+
+#[exception]
+fn HardFault(ef: &ExceptionFrame) -> ! {
+    panic!("{:#?}", ef);
+}
+
+#[exception]
+fn DefaultHandler(irqn: i16) {
+    panic!("Unhandled exception (IRQn = {})", irqn);
+}
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -27,13 +27,13 @@ fn main() -> ! {
     let mut pwr = p.PWR;
     let mut flash = p.FLASH.constrain();
     let mut rcc = p.RCC.constrain();
-    let backup_domain = rcc.cfgr.access_backup_domain(&mut pwr);
+    let backup_domain = rcc.backup_domain(&mut pwr);
     let clocks = rcc.cfgr.freeze(&mut flash.acr);
 
     let rtc = backup_domain.rtc(p.RTC, &clocks);
 
     loop {
-        writeln!(hstdout, "time: {}", rtc.read()).unwrap();
+        writeln!(hstdout, "time: {}", rtc.read_counts()).unwrap();
     }
 }
 
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -1,4 +1,4 @@
-//! Blinks an LED
+//! Outputs the current time in seconds to hstdout using the real time clock
 
 #![deny(unsafe_code)]
 #![deny(warnings)]
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -29,9 +29,8 @@ fn main() -> ! {
     let mut pwr = p.PWR;
     let mut rcc = p.RCC.constrain();
     let mut backup_domain = rcc.bkp.constrain(p.BKP, &mut rcc.apb1, &mut pwr);
-    let lse = backup_domain.enable_lse(rcc.lse);
 
-    let rtc = Rtc::rtc(p.RTC, lse, &mut backup_domain);
+    let rtc = Rtc::rtc(p.RTC, rcc.lse, &mut backup_domain);
 
     loop {
         writeln!(hstdout, "time: {}", rtc.seconds()).unwrap();
--- a/src/backup_domain.rs
+++ b/src/backup_domain.rs
@@ -1,10 +1,23 @@
-use stm32::{BKP};
+/*!
+  Registers that are not reset as long as Vbat or Vdd has power.
 
+  The registers retain their values during wakes from standby mode or system resets. They also
+  retain their value when Vdd is switched off as long as V_BAT is powered.
+
+  The backup domain also contains tamper protection and writes to it must be enabled in order
+  to use the real time clock (RTC).
+
+  Write access to the backup domain is enabled in RCC using the `rcc::Rcc::BKP::constrain()`
+  function.
+
+  Only the RTC functionality is currently implemented.
+*/
+
+use stm32::{BKP};
 
 /**
   The existence of this struct indicates that writing to the the backup
-  domain has been enabled. It is aquired by calling the `enable_backup_domain`
-  function on `rcc::CFGR`
+  domain has been enabled. It is aquired by calling `constrain` on `rcc::Rcc::BKP`
 */
 pub struct BackupDomain {
     pub(crate) _regs: BKP
--- a/src/backup_domain.rs
+++ b/src/backup_domain.rs
@@ -1,0 +1,18 @@
+use crate::rtc::Rtc;
+use crate::rcc::Clocks;
+
+use stm32::RTC;
+
+
+pub struct BackupDomain {
+    pub(crate) _0: ()
+}
+
+impl BackupDomain {
+    pub fn rtc(&self, regs: RTC, _clocks: Clocks) -> Rtc {
+        Rtc::rtc(regs)
+    }
+}
+
+
+
--- a/src/backup_domain.rs
+++ b/src/backup_domain.rs
@@ -4,13 +4,19 @@ use crate::rcc::Clocks;
 use stm32::RTC;
 
 
+/**
+  The existence of this struct indicates that writing to the the backup
+  domain has been enabled. It is aquired by calling the `enable_backup_domain`
+  function on `rcc::CFGR`
+*/
 pub struct BackupDomain {
     pub(crate) _0: ()
 }
 
 impl BackupDomain {
-    pub fn rtc(&self, regs: RTC, _clocks: Clocks) -> Rtc {
-        Rtc::rtc(regs)
+    /// Initialise the RTC
+    pub fn rtc(&self, regs: RTC, clocks: &Clocks) -> Rtc {
+        Rtc::rtc(regs, clocks)
     }
 }
 
--- a/src/backup_domain.rs
+++ b/src/backup_domain.rs
@@ -15,7 +15,8 @@
 
 use stm32::{BKP, rcc, RCC};
 
-use crate::rcc::Lse;
+use crate::rcc::{LSE};
+use time::Hertz;
 
 /**
   The existence of this struct indicates that writing to the the backup
@@ -28,6 +29,7 @@ pub struct BackupDomain {
 impl BackupDomain {
     // Enables the RTC device with the lse as the clock
     pub(crate) fn enable_rtc(&mut self, lse: Lse) {
+        // NOTE: Safe RCC access because we are only accessing bdcr
         let rcc = unsafe { &*RCC::ptr() };
         rcc.bdcr.modify(|_, w| {
             w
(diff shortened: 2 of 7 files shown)


TheZoq2:
After some consideration, I came to the conclusion that having a `BackupDomain` type that can be "created" by a function in the cfgr register is a better solution. From the `BackupDomain` type, an `Rtc` instance can be created.

While still a bit cumbersome, I think this is the best solution so far.

src/rcc.rs
```diff
@@ -143,6 +151,32 @@ impl CFGR {
         self
     }
 
+    /// Gives write access to the registers in the backup domain
+    pub fn enable_backup_domain(&mut self, pwr: &mut PWR) -> BackupDomain {
+        // NOTE: Access to apb1enr inside Rcc struct is means we have exclusive access
```
therealprof: Sentence doesn't parse for me

TheZoq2: :+1: That "is" was definitively not supposed to be there. I updated the wording for both comments.

I also changed the name of `enable_backup_domain` to `access_backup_domain` which is more descriptive of what's actually happening

examples/rtc.rs
```diff
@@ -0,0 +1,48 @@
+//! Blinks an LED
```
therealprof: This seems also wrong.

src/backup_domain.rs
```diff
@@ -0,0 +1,24 @@
...
+    pub(crate) _0: ()
+}
+
+impl BackupDomain {
+    /// Initialise the RTC
+    pub fn rtc(&self, regs: RTC, clocks: &Clocks) -> Rtc {
+        Rtc::rtc(regs, clocks)
+    }
+}
+
```
therealprof: Few too many newlines at the end of this file

therealprof:
Other than those nits, looks good to me. I don't really have any hard opinion on the initialisation question. It would also be great to have an a little more real world example for the use of an RTC like firing an alarm after a couple of seconds...

TeXitoi:
I'll have a look on it tomorrow.

TheZoq2:
Thanks, I pushed fixes for the style nits.

> It would also be great to have an a little more real world example for the use of an RTC like firing an alarm after a couple of seconds.

That's true, I'll try and write something. For reference, this is how I use it in my weather station https://github.com/TheZoq2/weather/blob/master/stmhardware/src/main.rs#L226

src/rcc.rs
```diff
@@ -143,6 +151,33 @@ impl CFGR {
         self
     }
 
+    /// Enables write access to the registers in the backup domain
+    pub fn access_backup_domain(&mut self, pwr: &mut PWR) -> BackupDomain {
+        // NOTE: Exclusive access to RCC because this struct is a member of
+        // Rcc
+        let apb1_enr = unsafe { &(*RCC::ptr()).apb1enr };
```
TeXitoi: The Rcc struct is public, thus it can be destructured. Better to add this method directly on Rcc, and you then should have access without unsafe to apb1enr by the APB1 struct

src/rcc.rs
```diff
@@ -143,6 +151,33 @@ impl CFGR {
         self
     }
 
+    /// Enables write access to the registers in the backup domain
+    pub fn access_backup_domain(&mut self, pwr: &mut PWR) -> BackupDomain {
```
TeXitoi: Maybe backup_domain is enough. Later, we can implement access to backup registers on this struct.

src/rcc.rs
```diff
@@ -143,6 +151,33 @@ impl CFGR {
...
+
+        // Enable access to the backup registers
+        pwr.cr.modify(|_r, w| {
+            w.
+                dbp().set_bit()
+        });
+
+
+        // Remember that the LSE needs to be initialised in clocks
+        self.backup_domain = Some(());
```
TeXitoi: I dont like that. backup_domain doesn't need any more clock, only RTC need.

src/rcc.rs
```diff
@@ -237,6 +272,20 @@ impl CFGR {
...
+            // in section 7.2.4. This gives a 32.768khz frequency for the RTC
+            rcc.bdcr.modify(|_, w| {
+                w
+                    // Enable external low speed oscilator
+                    .lseon().set_bit()
+                    // Enable the RTC
+                    .rtcen().set_bit()
+                    // Set the source of the RTC to LSE
+                    .rtcsel().lse()
+            });
```
TeXitoi: I'd prefer to have a `pub fn lse<F>(self, freq: F) -> Self` in CFGR (and a convenient `pub fn lse_is_32k(self)`), and then start it here and note the freq in `Clocks`

src/rtc.rs
```diff
@@ -0,0 +1,136 @@
...
+impl Rtc {
+    /**
+      Initialises the RTC, this should only be called if access to the backup
+      domain has been enabled
+    */
+    pub(crate) fn rtc(regs: RTC, _clocks: &Clocks) -> Self {
+        // Set the prescaler to make it count up once every second
+        // The manual on page 490 says that the prescaler value for this should be 7fffh
+        regs.prll.write(|w| unsafe{w.bits(0x7fff)});
+        regs.prlh.write(|w| unsafe{w.bits(0)});
```
TeXitoi: Can be calculated from the `Clocks` freq as

```rust
       let freq = 32768;
        let prl = freq - 1;
        assert!(prl < 1 << 20);
        rtc.modify(|s| {
            s.rtc.prlh.write(|w| unsafe { w.bits(prl >> 16) });
            s.rtc.prll.write(|w| unsafe { w.bits(prl as u16 as u32) });
        });
```

src/rtc.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    /// Starts an alarm which will trigger the RTCALARM interrupt in `time_seconds`
+    pub fn set_alarm(&mut self, time_seconds: u32) {
+        // Reset counter
+        self.perform_write(|s| {
+            s.regs.cnth.write(|w| unsafe{w.bits(0)});
+        });
+        self.perform_write(|s| {
+            s.regs.cntl.write(|w| unsafe{w.bits(0)});
+        });
```
TeXitoi: This in contrary to my usecase: I need to have cnt to be epoch for my alarm clock. For me, set alarm ask for a date, but must not change cnt.

src/rtc.rs
```diff
@@ -0,0 +1,136 @@
...
+        // Disable alarm interrupt
+        self.perform_write(|s| {
+            s.regs.crh.modify(|_, w| w.alrie().clear_bit());
+        })
+    }
+
+    /// Reads the current second since the RTC device was initialised
+    pub fn read(&self) -> u32 {
+        // Wait for the APB1 interface to be ready
+        while self.regs.crl.read().rsf().bit() == false {}
```
TeXitoi:
```rust
while !self.regs.crl.read().rsf().bit() {}
```

This pattern appear at other place. I know that's a personal preference, feel free to ignore if you don't like.

src/rtc.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Disables a previously set alarm
+    pub fn disable_alarm(&mut self) {
+        // Disable alarm interrupt
+        self.perform_write(|s| {
+            s.regs.crh.modify(|_, w| w.alrie().clear_bit());
+        })
+    }
+
+    /// Reads the current second since the RTC device was initialised
+    pub fn read(&self) -> u32 {
```
TeXitoi: I'd name it cnt or something like that, because we may also read the alarm time.

src/rtc.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    /// Enables the RTC second interrupt
+    pub fn listen_seconds(&mut self) {
+        self.perform_write(|s| {
+            s.regs.crh.modify(|_, w| w.secie().set_bit())
+        })
+    }
+
+    /// Disables the RTC second interrupt
+    pub fn unlisten_seconds(&mut self) {
```
TeXitoi: I'm not found of the set/disable for alarm and listen/unlisten on seconds, it might use the same naming convention as it listen for an event?

TeXitoi:
Also looks like we can't really separate BKP and RTC, thus maybe a unique Bkp object should take the backup registers and rtc functionnality. We must also be careful that only one Bkp object can be created, else we can have race with the modify dance.

TheZoq2:
Thanks for the thorough review!

TheZoq2:
Thanks for the thorough review

> For my use case, it lacks a `set_cnt` method.

That could be arranged, I just don't need it for my use case yet :)

A couple of your comments are related to clock initialisation which is a tricky problem.

The reason I chose the structure of `CFGR`->`BKP`-> `RTC` with clock initialisation being tied to BKP is that I don't want the RTC to exist without the relevant clock being initialised.

If we add a separate `lse` method, we need a way to ensure that it was started when the RTC is started. I suppose one method would be to have two functions such as 

```rust
/// Initialises backup domain but does not modify any clocks
backup_domain() -> BackupDomain<()>
/// Initialised backup domain and starts LSE with default frequency
/// Somehow statically encode that BKP is configured for RTC, perhaps with marker traits
backup_domain_rtc() -> BackupDomain
/// Sets the frequency of the LSE. If bkp_domain has been called, this changes the frequency
/// if not, the clock will be set as usual without affecting bkp
lse(...)
```

On paper, this structure sounds pretty reasonable in my opinion, any thoughts on it? If not I'll try to implement it this weekend

> Also looks like we can't really separate BKP and RTC, thus maybe a unique Bkp object should take the backup registers and rtc functionnality. We must also be careful that only one Bkp object can be created, else we can have race with the modify dance.

Initialisation has to be combined, but once they are created, they can be separated, right? That is unless we add a function to disable the RTC.

Also, unique Bkp objects are only a requirement if we add additional functionality to that struct, right? The current structure does not lead to any race conditions as far as I can tell. I'm not saying that's a permanent solution just trying to understand the problem

src/rcc.rs
TheZoq2: Good catch, it has been moved

src/rtc.rs
TheZoq2: Good point, that's probably a more general use case. I'll make it not reset the count and while i'm at it I'll add a `set_cnt` function

src/rtc.rs
TheZoq2: The difference is that listen_seconds/unlisten only turns on an interrupt while set_alarm also sets the counter value. Perhaps we want one function for setting the alarm and another to listen for events?

TeXitoi:
# The race problem

For the reader, any write in tthe backup domain must be surrounded by some `unlock_write` `commit_and_lock_write` (details in `perform_write` in the current implementation).

> Also, unique Bkp objects are only a requirement if we add additional functionality to that struct, right? The current structure does not lead to any race conditions as far as I can tell. I'm not saying that's a permanent solution just trying to understand the problem

current race:

```rust
    let bp1 = rcc.backup_domain(&mut pwr);
    let bp2 = rcc.backup_domain(&mut pwr);
    let clocks = rcc.cfgr.freeze(&mut flash.acr);

    let rtc1 = bp1.rtc(p.RTC, &clocks);
    let rtc2 = bp2.rtc(p.RTC, &clocks);

    // rtc1 and rtc2 in  2 different interrupts
    rtc1.set_alarm(1337);
    // -> unlock_write
    // -> interrupt
    {
        rtc2.set_cnt(42);
        // -> unlock_write (I suppose silent)
        // -> set cnt
        // -> commit_and_lock_write
    }
    // ->set alarm (not working as locked)
```

1. to fix this, we must ensure that not 2 `BackupDomain` can be constructed. I propose a `BKP` field in Rcc that have a `fn constrain(self) -> BackupDomain`
2. if we have a `BackupDomain` and a `Rtc`, and that we have write operations on them, we can have such races (writing and alarm and a backup register at the same time)

I can see 2 solutions.

# Solution 1: no static check for "RTC is started"

First, LSE is only used for RTC, this we can remote it from `Clocks`. I propose to add a `LSE` field in `Rcc` with a `fn freeze<F>(self, freq: F, &mut BackupDomain) ->LseClock` (and the helper `fn freeze_with_32k(self) ->LseClock`). The `&mut BackupDomain` is needed because the backup domain must be started before starting LSE. I would be a bit more logic to put `LSE` in `BackupDomain`, but then we would need to move `LSE` out of `BackupDomain`, and `BackupDomain` would not be useful anymore.

`struct LseClocks(Hertz)` would then only be available if LSE is started.

Our `BackupDomain` struct allow to do everything about RTC and backup registers. All the RTC method are usable without activating the RTC clock as far as I understand from the RM, but they will be quite useless as CNT will not be incremented.

We have on `BackupDomain` a method `fn start_rtc_with_lse(&mut self, Lse)` that start the RTC.

Advantages:
 - Safe (AFAIK)
 - simpler that solution 2

Drawbacks:
 - RTC can be used without starting the clock without any warning.
 - RTC can be started multiple times, which has no effect (not problematic now, but can be if we add later `start_rtc_with_hse` and friend)

# Solution 2: Static check of "RTC is started"

The solution is the same as solution 1 except that `BackupDomain` have a type parameter. By default, it is `()` (i.e. we have on BKP `fn constrain(self) -> BackupDomain<()>`). Access to the backup registers are implemented on `BackupDomain<T>`.

To start RTC, we use, on `BackupDomain<()>`, the method `fn start_rtc_with_lse(self, Lse) -> BackupDomain<Rtc>`. All the RTC metod are implemented on `BackupDomain<Rtc>`

Advantages:
 - Safe (I'm "more" sure than for solution 1)
 - RTC cannot be misused (AFAIK)

Drawbacks:
 - quite complex typestate
 - need generics if you want to use backup registers with and without RTC activated.

# Conclusion

What a tricky subject!

@TheZoq2 I think solution 2 ticks all your requirements, but not solution 1.

@therealprof @TheZoq2 what do you think?

src/rcc.rs
TeXitoi: Out of date because of my comment

src/rtc.rs
TeXitoi: I have a small preference for 3 methods then (`set_alarm`, `listen_alarm`, unlisten_alarm`) but nothing very strong. @therealprof what do you think?

TheZoq2:
It is indeed a tricky problem. I actually hadn't considered that you may want to write protect the backup domain after using it but I suppose there is a reason for that being possible :smile:. Your proposed design of having some sort of `perform_write` function sounds like it solves that problem.

Regarding the initialisation problem, as you say, solution 2 ticks all of my requirements so I would prefer it. It does add some extra complexity but i'd say that's better than having operations fail silently. If everyone else is happy with solution 2, I can try implementing it this weekend.

> current race:

That code wouldn't compile because there is just one `RTC` device which gets moved into the rtc struct when calling `bkp.rtc()`. We could (and probably should) do something similar with the backup domain to prevent similar problems when it gets extended.

TeXitoi:
> I actually hadn't considered that you may want to write protect the backup domain after using it

That's what you are doing [here](https://github.com/stm32-rs/stm32f1xx-hal/pull/17/files#diff-e5c4069fe6e355f0ba8dd9e1ab50a443R142) no?

>  If everyone else is happy with solution 2, I can try implementing it this weekend.

Great for me.

> That code wouldn't compile because there is just one RTC device which gets moved into the rtc struct when calling bkp.rtc()

You are right, my bad. It still apply if `BackupDomain` allow access to backup registers.

TheZoq2:
> That's what you are doing [here](https://github.com/stm32-rs/stm32f1xx-hal/pull/17/files#diff-e5c4069fe6e355f0ba8dd9e1ab50a443R142) no?

Oh yea, you're right. I thought you meant turning the backup domain write protection off and on

TeXitoi:
> Oh yea, you're right. I thought you meant turning the backup domain write protection off and on

Mmm, rereading the RM, I see that I've missunderstand something: I was thinking that the `perform_write` procedure was shared with the update of the backup registers, but that's not the case. Thus BackupDomain and Rtc can be separated. Proposition 1 and 2 became useless, we can have BackupDomain generate Rtc.

I rewrite the complete proposition according to that.

TeXitoi:
# Proposition 3

- `BKP` field in `Rcc` that have a `fn constrain(self) -> BackupDomain`
- `LSE` field in `Rcc` with a `fn freeze<F>(self, freq: F, &mut BackupDomain) ->LseClock` (and the helper `fn freeze_with_32k(self) ->LseClock`).
- `struct LseClocks(Hertz)` would then only be available if LSE is started.
- `BackupDomain` has a method `fn rtc_with_lse(&mut self, Lse, RTC) -> Rtc`

Advantages: simpler than proposition 2

Drawbacks: We can't implement properly the reset of the backup domain, as we would need the Rtc by value if created. Proposition 2 allow this easily.

therealprof:
@TeXitoi Sounds good. If in doubt you could always add a method taking a CS, to implement the reset of the Rtc without having the Rtc handle itself. Or we could add the registers to a singleton mega-struct...

TeXitoi:
@therealprof not sure I understand your proposition (CS?). The problem with backup reset is kenable backup -> enable rtc -> reset (or disable?) backup -> use rtc but you should not.

therealprof:
@TeXitoi I see. Disregard my comment about the CS then; I thought this was about resetting the RTC without having access to some other register. I'm not familiar with the backup domain and don't have time to look it up in the datasheet at the moment. I'm assuming we can't or don't want to separate it from the clocks?

TeXitoi:
In my 3 proposals, we don't touch to Clocks anymore.

Short description of the backup domain:
 - it consists of some register (that survive to standby and power off with vbat)
 - a tamper functionality (I don't know what it is).
 - the RTC

LSE is connected only to RTC, and is the only clock that survive to standby and power off with vbat. HSE/128 and LSI can also be used.

Backup domain must be started to be accessed. It can be reseted. Changing the RTC clock need a backup reset. Backup domain have a write protection feature (activated at startup) that can enable and disable writing in the backup and rtc registers.

TeXitoi:
All my propositions use the LSE tocken. But tour right that we can do proposition 2 without LSE token, enabling LSE in rtc_with_lse, accessing bdcr using unsafe as bdcr belong to the backup domain.

TeXitoi:
And proposition 2 have my preference to allow reset.

TeXitoi:
You should give apb1 and pwr to constrain on BKP

TeXitoi:
I know that it was my idea, but we must not impl on Rcc as this is a struct made to be restructured, and thus impl method will only work on Rcc before any constrain.

TheZoq2:
I went ahead and implemented a (slightly modified) solution 3 which seems pretty good, at least for my use case. Compared to solution 2, it clearly separates `RTC` from `BKP` which seems like a good since they aren't really coupled after initialisation.

> And proposition 2 have my preference to allow reset.

What kind of reset are you considering. As far as I can tell, there is no such feature in the `BKP`. If there is, the reset function could consume the RTC again in order to ensure that it isn't affected.

> `LSE` field in `Rcc` with a `fn freeze<F>(self, freq: F, &mut BackupDomain) ->LseClock` (and the helper `fn freeze_with_32k(self) ->LseClock`)

It turns out that without hooking up an external oscilator, the frequency of `LSE` can't be changed. I made freeze set it to 32khz and will put arbitrary frequencies off to the future if I need them and have an oscilator to test it with.

Another note:

Do we want to call functions for manipulating the alarm `<>_counts` or `<>_seconds`? The current sets the prescaler to always use seconds so calling them seconds makes sense. On the other hand, we could change that in the future which would require changing the function names. I suppose that might also be a good thing because code would break if we change the prescaler while keeping the same functions.

TheZoq2:
Also, we now have both `rcc::BKP` and `stm32::BKP`, do we need/want that and do we want to rename rcc::BKP to something that doesn't conflict with the PAC name?

TheZoq2:
I pushed some documentation changes along with some more rtc functions:

- Module level comments for `rtc` and `backup_domain`
- Update doc comments to reflect current implementation
- A new example: `blinky_rtc` which showcases alarm usage

In order to implement blinky_rtc I had to add a `wait_alarm` function which is similar to `wait` for normal timers. I also made `set_alarm` reset clear the alarm flag which I *think* is the desired behaviour.

I also went ahead and changed `counts` to `seconds` wherever that is used since in the current implementation, `counts == seconds`

therealprof:
@TheZoq2 I'm generally okay with this and leave the final word to @TeXitoi . Please also add an entry to  CHANGELOG.md .

TheZoq2:
Changelog entry has been added :)

src/rcc.rs
```diff
@@ -318,6 +330,59 @@ impl CFGR {
...
+        let rcc = unsafe { &*RCC::ptr() };
+
+        rcc.bdcr.modify(|_, w| {
+            w
+                // Enable external low speed oscilator
+                .lseon().set_bit()
+                // Enable the RTC
+                .rtcen().set_bit()
+                // Set the source of the RTC to LSE
+                .rtcsel().lse()
```
TeXitoi: That's a bit strange to start RTC here. I think we can simplify the LSE thing by just starting it in the RTC constructor.

src/rcc.rs
```diff
@@ -318,6 +330,59 @@ impl CFGR {
...
+        });
+
+        // NOTE: Chosing another frequency requires an external oscilator
+        Lse{freq: Hertz(32768)}
```
TeXitoi: No, starting LSE need an external oscilator, LSE is for LowSpeed External (oscilator). It's for example present on the blue pill (the  black rectangular thing) but not on the black pill. Having something different from a 32k oscilator must be anecdotal.

src/rcc.rs
```diff
@@ -318,6 +330,59 @@ impl CFGR {
...
+        apb1.enr().modify(|_r, w| {
+            w
+                .bkpen().set_bit()
+                .pwren().set_bit()
+        });
+
+        // Enable access to the backup registers
+        pwr.cr.modify(|_r, w| {
+            w.
+                dbp().set_bit()
```
TeXitoi: strange `.` placement.

examples/rtc.rs
```diff
@@ -0,0 +1,49 @@
...
+    let mut hstdout = hio::hstdout().unwrap();
+
+    let p = hal::stm32::Peripherals::take().unwrap();
+
+    let mut pwr = p.PWR;
+    let mut rcc = p.RCC.constrain();
+    let backup_domain = rcc.bkp.constrain(p.BKP, &mut rcc.apb1, &mut pwr);
+    let lse = rcc.lse.freeze(&backup_domain);
+
+    let rtc = Rtc::rtc(p.RTC, lse, &backup_domain);
```
TeXitoi: Maybe it would be simpler to have the constructor on backup_domain to do `backup_domain.rtc(p.RTC, lse);`

src/rtc.rs
```diff
@@ -0,0 +1,180 @@
...
+
+
+impl Rtc {
+    /**
+      Initialises the RTC. The `Lse` and `BackupDomain` structs are created by
+      `Rcc.lse.freeze()` and `Rcc.bkp.constrain()` respectively
+    */
+    pub fn rtc(regs: RTC, lse: Lse, _bkp: &BackupDomain) -> Self {
+        // Set the prescaler to make it count up once every second
+        // The manual on page 490 says that the prescaler value for this should be 7fffh
```
TeXitoi: unsync comment.

src/rtc.rs
```diff
@@ -0,0 +1,180 @@
...
+        result
+    }
+
+    /// Set the current rtc value to the specified amount of seconds
+    pub fn set_seconds(&mut self, seconds: u32) {
+        self.perform_write(|s| {
+            s.regs.cnth.write(|w| unsafe{w.bits(seconds >> 16)});
+        });
+        self.perform_write(|s| {
+            s.regs.cntl.write(|w| unsafe{w.bits(seconds)});
```
TeXitoi: You can do these 2 writes in the same perform_write.

src/rtc.rs
```diff
@@ -0,0 +1,180 @@
...
+    */
+    pub fn set_alarm(&mut self, seconds: u32) {
+        // Set alarm time
+        // See section 18.3.5 for explanation
+        let alarm_value = seconds - 1;
+        self.perform_write(|s| {
+            s.regs.alrh.write(|w| unsafe{w.alrh().bits((alarm_value >> 16) as u16)});
+        });
+        self.perform_write(|s| {
+            s.regs.alrl.write(|w| unsafe{w.alrl().bits((alarm_value & 0x0000ffff) as u16)});
```
TeXitoi:
```rust
s.regs.alrl.write(|w| unsafe{w.alrl().bits(alarm_value as u16)});
```
is enough, `as u16` will truncate as needed.

And you can do these 2 writes in the same perform_write.

src/rtc.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// Disables the RTCALARM interrupt
+    pub fn unlisten_alarm(&mut self) {
+        // Disable alarm interrupt
+        self.perform_write(|s| {
+            s.regs.crh.modify(|_, w| w.alrie().clear_bit());
+        })
+    }
+
+    /// Reads the current time
+    pub fn read_seconds(&self) -> u32 {
```
TeXitoi: https://rust-lang-nursery.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

Getters don't have `get_` or `read_` prefix. I thus propose to rename this `fn seconds(&self) -> u32`

src/rtc.rs
```diff
@@ -0,0 +1,180 @@
...
+    _lse: Lse,
+}
+
+
+impl Rtc {
+    /**
+      Initialises the RTC. The `Lse` and `BackupDomain` structs are created by
+      `Rcc.lse.freeze()` and `Rcc.bkp.constrain()` respectively
+    */
+    pub fn rtc(regs: RTC, lse: Lse, _bkp: &BackupDomain) -> Self {
```
TeXitoi: Better to take a `&mut BackupDomain`

src/rtc.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub fn clear_alarm_flag(&mut self) {
+        self.perform_write(|s| {
+            s.regs.crl.modify(|_, w| w.alrf().clear_bit())
+        })
+    }
+
+    /**
+      Return `Ok(())` if the alarm flag is set, `Err(nb::WouldBlock)` otherwise.
+
+      **Note**: Does not clear the alarm flag
```
TeXitoi: Any reason of this? I'm OK with this, just curious.

src/rtc.rs
TheZoq2: Why? As far as I know, the backup domain can't affect the RTC apart from enabling access. I'm only taking it as a token to verify that it is enabled.

src/rtc.rs
TheZoq2: I thought that was the way it is done in the general purpose timers but I guess it's not. I'll make it clear the flag.

TheZoq2:
Good comments, i'll fix them when I get home from work

src/rtc.rs
TeXitoi: Because in my mind, `rcc.bdcr.modify(|_, w| w.rtcen().set_bit())` should be in this method, and this is a write operation. But maybe I'm wrong?

examples/rtc.rs
TheZoq2: This is more similar to what is done for every other device in the crate. Do we have any reason to do it differently for the RTC?

src/rcc.rs
TheZoq2: That requires leaking access to BDCR outside RCC, do we want to do that?

examples/rtc.rs
TeXitoi: No, if that's more like the other peripherals, you're right.

src/rcc.rs
TeXitoi: As BDCR is Backup Domain Control Register, as long as we have access to BackupDomain, that doesn't chock me. But that's right that starting LSE without starting RTC is useless.

src/rcc.rs
TheZoq2: I thought so too at first, but this is what the documentation says in section 8.2.4:

> The LSE crystal is a 32.768 kHz Low Speed External crystal or ceramic resonator. It has the advantage providing a low-power but highly accurate clock source to the real-time clock peripheral (RTC) for clock/calendar or other timing functions.
> 
> The LSE crystal is switched on and off using the LSEON bit in Backup domain control register (RCC_BDCR)

TheZoq2:
Alright, I have fixed the issues you pointed out

I decided to move Lse into backup_domain because backup_domain was a bit too coupled to `rcc`. I'm not 100% happy with it because I prefer the LSE.freeze() over `enable_lse(LSE)` but I also dislike the pub(crate) that it required with the current design. What are your thoughts on that?

TeXitoi:
I don't have any preference between the 2. We can also just remove the LSE token an just start LSE in the Rtc constructor.

src/rtc.rs
TeXitoi: You forgot to merge these writes.

src/rcc.rs
```diff
@@ -1,11 +1,17 @@
...
+use stm32::{self, rcc, RCC, PWR};
 
 use flash::ACR;
 use time::Hertz;
 
+use backup_domain::BackupDomain;
+
+pub struct BackupDomainEnabledToken {
+    _0: ()
+}
```
TeXitoi: Unused struct.

TheZoq2:
> I don't have any preference between the 2. We can also just remove the LSE token and just start LSE in the Rtc constructor.

Getting rid of the LSE token sounds like a good idea, the reason I didn't do that before was that I thought it is used for something else as well, but that's wrong. I'll fix that this evening

TheZoq2:
There, the changes have been commited

TeXitoi:
Working here on rusty clock! Thanks!

PR #19, opened 2019-02-16
Add support for the ADC preripheral

TheZoq2:
This is another port of a previous PR from the old repo https://github.com/japaric/stm32f103xx-hal/pull/114

The implementation is basically the same, with the only change being an added error type which is returned if a conversion on another channel is already happening when a read is started. As far as I know, this should fix all outstanding issues.

It would be nice to add static checks for this, but as mentioned in the previous PR, that would require changes to `embedded_hal`. I therefore propose that we merge this implementation and look into doing the check statically in the future.

CC: @tib888 This might be of interest to you, you had some very insightful comments in the previous PR

### DIFF
--- a/examples/adc.rs
+++ b/examples/adc.rs
@@ -1,6 +1,7 @@
 //! Makes an analog reading on channel 0 and prints it to itm
 
 #![deny(unsafe_code)]
+#![deny(warnings)]
 #![no_main]
 #![no_std]
 
@@ -24,7 +25,6 @@ use stm32f1xx_hal::prelude::*;
 use rt::ExceptionFrame;
 use stm32f1xx_hal::adc::{self};
 use embedded_hal::adc::OneShot;
-use cortex_m::asm;
 
 #[entry]
 fn main() -> ! {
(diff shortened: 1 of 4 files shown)


therealprof:
@TheZoq2 We do have an (IMHO) very nice Adc implementation in [the STM32F0xx-HAL](https://github.com/stm32-rs/stm32f0xx-hal/blob/master/src/adc.rs) which feels a bit more modern to me. I'm not sure how compatible the approaches (I think they have different Adc generations) but maybe a few bits can be applied to this driver as well?

geomatsi:
Hi @TheZoq2 
Do you have any plans to move forward with ADC support for stm32f1 ? If yes, then I can help with testing, just CC me in your PRs. If not, then I am going to take your code as a baseline and try to rework following to stm32f0xx-hal model.

Regards,
Sergey

TheZoq2:
@geomatsi I was planning on getting to it at some point, but other projects and features have taken priority. If you want to take over the work on this, feel free :)

TheZoq2:
Closing this in favour of #51

PR #20, opened 2019-02-16
stm32f100 - compiles, but PWM broken

Harnesser:
Hi there!

This is a WIP pull request to support the STM32F100 which is on the valueline discovery board.

It looks like not all chips supported by `stm32f1` are supported by this HAL wrapper. Maybe they should be added to the `features` list in `Cargo.toml` as they become supported? 

I've added a `stm32f100` feature. PWM is disabled for the moment until I figure out how to fix the `.ccr()`-not-found problem.

A version of  `led.rs` on PC8 (with an `.set_high()` call that was missing) was tested and working. I'm going to have a look at timers and delays next. The ultimate goal is to be able to use this crate along with https://github.com/JohnDoneth/hd44780-driver

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,21 @@ cargo build --features stm32f103
 
 If no device is specified, the crate does not compile.
 
+Alternatively, adding the following to your project's `Cargo.toml` will target
+your chip without needing the `--features` argument:
+
+```
+[features]
+default = ["stm32f100", "rt"]
+rt = ["stm32f1xx-hal/rt"]
+stm32f100 = ["stm32f1xx-hal/stm32f100"]
+```
+
+## Supported Microcontrollers
+
+* STM32F100
+* STM32F103
+
 ## License
 
 Licensed under either of
--- a/README.md
+++ b/README.md
@@ -10,20 +10,26 @@ This crate will eventually contain support for multiple microcontrollers in the
 stm32f1 family. Which specific microcontroller you want to build for has to be
 specified with a feature, for example `stm32f103`.
 
-```
-cargo build --features stm32f103
-```
+If no microcontroller is specified, the crate will not compile.
 
-If no device is specified, the crate does not compile.
+### Building an Example
 
-Alternatively, adding the following to your project's `Cargo.toml` will target
-your chip without needing the `--features` argument:
+If you are compiling the crate on its own for development or running examples, 
+specify your microcontroller on the command line. For example:
 
 ```
-[features]
-default = ["stm32f100", "rt"]
-rt = ["stm32f1xx-hal/rt"]
-stm32f100 = ["stm32f1xx-hal/stm32f100"]
+cargo build --features stm32f103 --example led
+```
+
+### Using as a Dependency
+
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -16,6 +16,21 @@ cargo build --features stm32f103
 
 If no device is specified, the crate does not compile.
 
+Alternatively, adding the following to your project's `Cargo.toml` will target
+your chip without needing the `--features` argument:
+
+```
+[features]
```
therealprof: The same can be achieved more easily by simply enabling the features in the dependency section:
e.g.
```
[dependencies.stm32f1xx-hal]
features = ["stm32f100", "rt"]
```

No need to declare new features in the using crate and forward them.

therealprof:
> I've added a `stm32f100` feature. PWM is disabled for the moment until I figure out how to fix the `.ccr()`-not-found problem.

That may require a patch (and new release of stm32-rs).

Other than the nit in the README.md this looks great.

src/rcc.rs
```diff
@@ -261,6 +262,7 @@ impl CFGR {
         }
 
         // set prescalers and clock source
+        #[cfg(not(feature = "stm32f100"))]
```
TheZoq2: Maybe it makes more sense to explicitly specify the boards which things do work on, i.e. replace `not(feature = "stm32f100")` with `feature = "stm32f103"`. Though I suppose this depends on how specific these things are to different boards.

@therealprof @TeXitoi Any thoughts on this? How is it done in the other HALs?

therealprof: In other hal impls we're doing it exactly like you suggest, however cleaning that up shouldn't be a big deal. But now that you mention it, there's another thing missing... CI.

README.md
```diff
@@ -16,6 +16,21 @@ cargo build --features stm32f103
 
 If no device is specified, the crate does not compile.
 
+Alternatively, adding the following to your project's `Cargo.toml` will target
```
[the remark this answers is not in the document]
therealprof: > This is correct, but AFAIK, `cargo build --features`

Should work just fine by doing `cargo build --features="stm32f1xx-hal/stm32f100"` but unless your crate can target different MCUs I wouldn't recommend this.

src/rcc.rs
Harnesser: Makes sense. Updated.

README.md
Harnesser: Yeah, that's a far better idea. Updated.

README.md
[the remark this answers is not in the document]
Harnesser: Had another go at the README. Added 2 sub-sections to the Usage section for dev/examples and use as a dependency

Harnesser:
I've no idea what has to happen to .travis.yml - any suggestions?

therealprof:
> I've no idea what has to happen to .travis.yml - any suggestions?

Search for 103, add a similar line for 100 right below it. ;)

therealprof:
Nice, compiles fine. Thanks a lot.

PR #21, opened 2019-02-17
Added link to docs.rs to README

jounathaen:
I think, a link to the docs does not hurt in the readme ;-)


TheZoq2:
Thanks for the PR! This is a good idea, though I would prefer a link to just docs.rs/stm32f1xx_hal rather than to the current version. That way we don't have to remember to update it each release :)

jounathaen:
Yes, of course. I changed that.

Issue #22, opened 2019-02-17
Improve Documentation


jounathaen:
The documentation "has a little room for improvments"
:-D

Especially in terms of "How do I use XY"

TheZoq2:
That is definitely true, we should probably work on that! I know most if it is lacking, but is there any area you feel is especially lacking?

TheZoq2:
Closing in favour of #158

PR #23, opened 2019-02-17
Documentation

jounathaen:
I tried to improve the documentation a bit. (#22)

Please review, as I'm new to this crate and I might have misunderstood things.
The example is adapted from the serial example.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -1,6 +1,41 @@
 //! # Serial Communication
 //! This module contains the functions to utilize the USART (Universal
 //! synchronous asynchronous receiver transmitter (USART)
+//!
+//!
+//! ## Example usage:
+//! ```rust
+//!// prelude: create handles to the peripherals and registers
+//!let p = stm32::Peripherals::take().unwrap();
+//!let cp = cortex_m::Peripherals::take().unwrap();
+//!let mut flash = p.FLASH.constrain();
+//!let mut rcc = p.RCC.constrain();
+//!let clocks = rcc.cfgr.freeze(&mut flash.acr);
+//!let mut afio = p.AFIO.constrain(&mut rcc.apb2);
+//!let mut gpioa = p.GPIOA.split(&mut rcc.apb2);
+//!
+//!// USART1 on Pins A9 and A10
+//!let pin_tx = gpioa.pa9.into_alternate_push_pull(&mut gpioa.crh);
+//!let pin_rx = gpioa.pa10;
+//!// Create an interface struct for USART1 with 9600 Baud
+//!let serial = Serial::usart1(
+//!    p.USART1,
+//!    (pin_tx, pin_rx),
+//!    &mut afio.mapr,
+//!    9_600.bps(),
+//!    clocks,
+//!    &mut rcc.apb2,
+//!);
+//!
+//!// separate into tx and rx channels
+//!let (mut tx, mut rx) = serial.split();
+//!
+//!// Write 'R' to the USART
+//!block!(tx.write(b'R')).ok();
+//!// Receive a byte from the USART and store it in "received"
+//!let received = block!(rx.read()).unwrap();
+//! ```
+
 use core::marker::PhantomData;
 use core::ptr;
 use core::sync::atomic::{self, Ordering};
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -188,8 +188,9 @@ macro_rules! hal {
                     Serial { usart, pins }
                 }
 
-                /// Starts listening by enabling _RX buffer not empty_ interrupt and _TX register
-                /// empty_ interrupt
+                /// Starts listening to the USART by enabling the _Received data
+                /// ready to be read (RXNE)_ interrupt and _Transmit data
+                /// register empty (TXE)_ interrupt
                 pub fn listen(&mut self, event: Event) {
                     match event {
                         Event::Rxne => self.usart.cr1.modify(|_, w| w.rxneie().set_bit()),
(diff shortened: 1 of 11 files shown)


src/serial.rs
```diff
@@ -101,7 +140,13 @@ macro_rules! hal {
         ),
     )+) => {
         $(
+            $(#[$meta])*
             impl<PINS> Serial<$USARTX, PINS> {
+
+            /// Configures the Serial interface and creates an interface struct.
+            /// `baud_rate` configures the baud rate of the interface. The other parameters are
+            /// handles to the corresponding registers, to make Rust's borrow checker happy.
```
TheZoq2: The regs aren't passed to make the borrow checker happy, they are passed to be modified in order to make the device work.

TheZoq2: The regs aren't passed to make the borrow checker happy, they are passed to be modified in order to make the device work.

src/serial.rs
```diff
@@ -101,7 +140,13 @@ macro_rules! hal {
...
+            /// (Ownership must be passed to this function)
```
TheZoq2: I would prefer "Takes ownership of the usart device" or something along those lines

src/serial.rs
```diff
@@ -141,24 +186,31 @@ macro_rules! hal {
                     Serial { usart, pins }
                 }
 
+                /// Starts listening by enabling _RX buffer not empty_ interrupt and _TX register
+                /// empty_ interrupt
```
TheZoq2: "RX buffer not empty" does not sound right. The interrupts are just "tx and rx interrupts"

src/serial.rs
```diff
@@ -403,6 +458,7 @@ hal! {
         |remap| remap == 1,
         APB2
     ),
+    #[doc(hidden)]
```
TheZoq2: I don't like hiding the docs for the other devices. A comment similar to that for USART1 would be better imo

src/serial.rs
```diff
@@ -1,3 +1,41 @@
+//! # Serial Communication
+//! This module contains the functions to utilize the USART (Universal
+//! synchronous asynchronous receiver transmitter (USART)
```
eldruin:
```suggestion
//! synchronous asynchronous receiver transmitter)
```
Small duplication here.

src/serial.rs
```diff
@@ -1,3 +1,41 @@
...
+//!
+//!
+//! ## Example usage:
+//! ```rust
+//!// prelude: create handles to the peripherals and registers
```
eldruin: I would add at least one space between //! and the code for reading ease when reading the documentation un-rendered.

src/serial.rs
jounathaen: Yes, but without the borrow checker, the registers could be encoded in the function statically (C-style), or have I misunderstood the concept of the peripheral struct?

I thought, that this is the reason that 6 parameters are passed to that function, whilst only one is a real parameter.

jounathaen: (But yes, that is sloppy language)

src/serial.rs
jounathaen: I now use the names from the datasheet:
- Transmit data register empty
- Received data ready to be read

src/serial.rs
```diff
@@ -101,7 +140,15 @@ macro_rules! hal {
...
         $(
+            $(#[$meta])*
+            /// The behaviour of the functions is equal for all three USARTs.
+            /// Except that they are using the corresponding USART hardware and pins.
             impl<PINS> Serial<$USARTX, PINS> {
+
+                /// Configures the Serial interface and creates an interface struct.
+                /// `baud_rate` configures the baud rate of the interface. The other parameters are
+                /// handles to the corresponding registers which must be changed by this function.
+                /// This function takes ownership of these registers.
```
TheZoq2: This isn't 100% correct for two reasons:

- Not all other parameters are registers, `Clocks` is a struct which 1. indicates that the clock configuration is frozen and 2, gives the frequencies of the various clocks. Also, pins are not registers, they are input and output pins
- It only takes ownership of the pins and usart device registers, the other registers are passed by reference in order to initialise the device

TheZoq2:
Thanks for the updates! 

Apart from the serial init comment, it looks good to me now

src/serial.rs
jounathaen: Thanks for the clarification. I adapted it accordingly to your explanation and I learned a lot about the functionality :+1:

TheZoq2: Looks almost perfect now!

> The `serial` struct takes ownership of the `PINS` and `USARTX` device registers.

I read this as "takes ownership of the `PINS` device registers and `USARTX` device registers which is incorrect, `PINS` is a tuple of individual pins. Perhaps changing it to "Takes ownership over the `USARTX` device registers and the specified `PINS`" would make it more clear.

> The `serial` struct does not take ownership of `MAPR` and `APBX`

Another, more style related nitpick is that I'm not sure if this comment is required as that can be infered from them being passed as references. 

Other than those two things, it looks good to me

jounathaen: As you wish, master!

TheZoq2:
Looks perfect now, great work! And thanks for being patient with my nitpicks :)

@therealprof I'd say this is ready to merge. Just to be sure since I've never done this before, do I do a merge or a squash? It looks like #20 was squashed but I can't tell for sure

therealprof:
@TheZoq2 Squashing sounds like a good idea. I'd only keep the individual commits if they make sense on their own but these are mostly fixes on top of other commits.

Issue #26, opened 2019-02-20
LED example toggle behavior


alx741:
The LED example turns it on (PC13) briefly and then turns it back off. It happens in this line:

https://github.com/stm32-rs/stm32f1xx-hal/blob/37756dbd0fed43e3175f8c54f0e901b88297cc7f/examples/led.rs#L27

Stepping through it,  it seems to be writing to BSRR (a couple steps after the LED was turned on) turning it off.

Is this a bug or am I missing something?

`Cargo.toml`

```
[dependencies]
cortex-m = "0.5.8"
cortex-m-rt = "0.6.5"
cortex-m-semihosting = "0.3.2"
panic-halt = "0.2.0"
panic-semihosting = "0.5.1"

[dependencies.stm32f1xx-hal]
version = "0.2.0"
features = ["stm32f103", "rt"]
```

TheZoq2:
Looks like a bug to me, a recent change changed that line from `into_push_pull_output()` to `into_push_pull_output().set_high()` but the LED is active low.

alx741:
ah of course! this is meant for the bluepill board and so the pin must be low in order for the LED to be on

PR #27, opened 2019-02-20
Fix broken LED example (#26)

TheZoq2:
Fix for #26 

This seems to have been broken in #20 where the affected line was changed from `gpioc.pc13.into_push_pull_output(&mut gpioc.crh);` to `gpioc.pc13.into_push_pull_output(&mut gpioc.crh).set_high();`

I think explicitly calling `set_low` is better than the previous case since it's more explicit and as far as I know, we don't specify wether or not pins will be high or low after `into_push_pull_output`. Or was there a reason `set_low` wasn't explicitly called?

### DIFF
--- a/examples/blinky.rs
+++ b/examples/blinky.rs
@@ -1,4 +1,9 @@
 //! Blinks an LED
+//!
+//! This assumes that a LED is connected to pc13 as is the case on the blue pill board.
+//!
+//! Note: Without additional hardware, PC13 should not be used to drive an LED, see page 5.1.2 of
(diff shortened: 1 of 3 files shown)


therealprof:
The only (very minor) issue I have with this change is that there's a certain assumption about the utilised hardware in here; what works for you will break for others (either because there's no LED at the utilised pins or it has the polarity the other way around). I think for a generic crate we should try to stay as generic as possible (i.e. make few or better no assumptions) and instead add commentary about all taken assumptions and what the code would do if the assumptions hold.

TheZoq2:
Is this roughly what you had in mind? @therealprof

PR #29, opened 2019-02-21
Edition 2018

### LINKED ISSUE #16: Exporting the sdv2rust crate as device
For consistency, I propose to re export the sdv2rust crate as `device`. It is coherent with the name used by RTFM: https://japaric.github.io/cortex-m-rtfm/book/by-example/app.html

If there is approvals and no veto, I can propose such a PR.

burrbull:
2018 edition path style.

Closes #16 .

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ features = ["unproven"]
 version = "0.2.2"
 
 [dev-dependencies]
+panic-halt = "0.2.0"
 panic-semihosting = "0.5.1"
 panic-itm = "0.4.0"
 # cortex-m-rtfm = "0.3.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ license = "MIT OR Apache-2.0"
 name = "stm32f1xx-hal"
 repository = "https://github.com/stm32-rs/stm32f1xx-hal"
 documentation = "https://docs.rs/stm32f1xx-hal"
+edition = "2018"
 version = "0.2.0"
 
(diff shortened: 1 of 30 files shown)


TheZoq2:
Thanks for the PR, this is definitely a good idea.

The changes look good to me, but the examples also have to be updated to the new edition, would you be able to do that?

Also, I don't think we came to a conclusion about what name to use for the "pac" crate in #16 . Perhaps we should stick with stm32 until there is an "official" conclusion as we would otherwise have 2 breaking changes.

burrbull:
As you can see, I've exported all three (`stm32`, `device` and `pac`) names. I prefer `device`, but `pac` is shorter.

TheZoq2:
Ah right, I missed that. In that case, the PR looks good apart from the missing examples

burrbull:
I am working on examples.

burrbull:
Done.

TheZoq2:
Looks almost perfect, though I do have some questions:

- Why did you replace `panic_semihosting` with `panic_halt`, i'm fine with either but I am curious
- What does the panic example show? The description and code look like it just shows printing hello world, but then the name panic doesn't make sense

burrbull:
I think examples should be as clean as possible. So I deleted panic handlers from most of examples and created additional example with panic illustration. Furthermore semihosting works only with debugger.

therealprof:
@burrbull I agree. Thanks for caring about the details.

TheZoq2:
Sounds reasonable, I'd say this is ready to merge then. Thanks for the help!

PR #30, opened 2019-02-22
Fix mixed up USART clocks

alexandermorozov:
For both stm32f100 and stm32f103 families USART1 is clocked by PCLK2, and all other USARTS by PCLK1.

I've checked ref manuals for both families. And it fixes baud rate on my f103.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -468,7 +468,7 @@ hal! {
         usart1en,
         usart1rst,
         usart1_remap,
-        pclk1,
+        pclk2,
         bit,
         |remap| remap == 1,
         APB2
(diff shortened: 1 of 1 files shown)


TeXitoi:
@gbip As you've done the modifications, does it broke your usecase?

alexandermorozov:
And RM 6.3.8 p.92 for stm32f100. I couldn't find explicit mention in USART section, but bits in clock gating registers confirm this layout.

gbip:
I can't believe I inverted all of them, I apologize.
I checked in the ref manual at first, but I had troubles with git and things got a little bit messy between various branches...

PR #31, opened 2019-02-24
Restore Dma transfer support

burrbull:


nraynaud:
May I suggest using this opportunity to modify `dma::CircBuffer::peek()`?
 - The name `peek()` suggest that the state is observed without being changed, but in this implementation it resets an essential bit that will allow the DMA to wrap around and overwrite the buffer
 - When we want to peek, we want to have a read only access to the entire buffer, not only the half the DMA is currently not writing. In particular, the message received from a serial device reader might be smaller than one half of the buffer, so the DMA is still pointing into this half, and we want to read the buffer anyways (imagine receiving 'Ok' into a 16 bytes buffer).
 - We might want to rename `peek()` to a real access operation, with a clear semantic that the user is given access to the half for the last time, because the DMA will overwrite it.

Cargo.toml
```diff
@@ -17,7 +17,7 @@ features = ["stm32f103", "rt"]
 cortex-m = "0.5.8"
 nb = "0.1.1"
 cortex-m-rt = "0.6.7"
-stm32f1 = "0.6.0"
+stm32f1 = "0.7.0"
```
therealprof: Uhm, can I borrow your crystal ball for a sec? 😅

jamwaffles:
> When we want to peek, we want to have a read only access to the entire buffer, not only the half the DMA is currently not writing. In particular, the message received from a serial device reader might be smaller than one half of the buffer, so the DMA is still pointing into this half, and we want to read the buffer anyways (imagine receiving 'Ok' into a 16 bytes buffer).

I'd really like to see this as well - it would make implementing a serial AT driver for those little ESP8266 modules _much_ easier.

Cargo.toml
burrbull: Of course not. You can break it. The art of seeing the future is passed down from generation to generation.

PS. See https://github.com/stm32-rs/stm32-rs/pull/116

burrbull: For testing you can use https://github.com/burrbull/stm32f1 until 0.7.0 be published.

burrbull:
The sense of a circular buffer is that you shall not have access to the part of the buffer that is being written to.

>     * (imagine receiving 'Ok' into a 16 bytes buffer).
What you mean saying 16 bytes buffer. Maybe you mean buffer of 16-bits values?

nraynaud:
I mean that you don't know what message you will receive, you start your UART with a circular buffer of 16 bytes. Suddenly "Ok" arrives, and the UART line goes idle.

How do you go read what's in the buffer?

burrbull:
Nohow. You have to wait next receiving until DMA HalfTransfer or TransferComplete interrupt arise and free buffer.
https://github.com/japaric/stm32f103xx-hal/blob/master/examples/reactive-serial-circ.rs

This is Rust safety system.

nraynaud:
1) There is no safety violation in having a reader and a writer going in the same direction as long as the writer doesn't wrap (get the writer pointer before the read and check the writer overflow after the read)

2) If there were a concern, then you need a method to lift the pen of the writer (and maybe tell it to go writing somewhere else while you read a partial half).

burrbull:
> 2. If there were a concern, then you need a method to lift the pen of the writer (and maybe tell it to go writing somewhere else while you read a partial half).

Sounds still dangerous. Maybe you have code to demonstrate?

In any case such suggestions should be discussed on embedded-hal where are more people and devices.

burrbull:
I'll make new.

PR #37, opened 2019-03-06
Improve example documentation

TheZoq2:
As a step towards closing #37 I want to start improving the documentation of this crate and I think the examples are a good place to start. I added a bunch of comments to the blinky and serial examples which try to explain what each line in the example does.

I also improved the main crate documentation page by adding the blinky example to it and clarifying how to specify the crate as a dependency. I also kind of think we should do that to the readme since that's where people will first encounter the crate, is that a good idea?

I intend to give the rest of the examples similar treatment, but I would like some feedback on the two examples that are already done so I don't make the same misstakes in *all* of them :). I'm not sure if we should make all of them this verbose, especially for the things that are the same between all of them. What do you think?

PS: Turns out pressing enter when creating a PR sends the PR without warning which is why it didn't have a description at first :sweat_smile:

### DIFF
--- a/examples/blinky.rs
+++ b/examples/blinky.rs
@@ -5,6 +5,13 @@
 //! Note: Without additional hardware, PC13 should not be used to drive an LED, see page 5.1.2 of
 //! the reference manaual for an explanation. This is not an issue on the blue pill.
 
+//! This example assumes that a LED is connected to PC13 which is where an LED is connected on the
+//! [blue_pill] board. If you have different hardware, you need to change which gpio pin is used.
+//!
+//! Note: PC13 can not source much current and should not be used to drive an LED directly.
+//!
+//! [blue_pill]: http://wiki.stm32duino.com/index.php?title=Blue_Pill
+
 #![deny(unsafe_code)]
 #![deny(warnings)]
 #![no_std]
--- a/examples/blinky.rs
+++ b/examples/blinky.rs
@@ -47,12 +47,14 @@ fn main() -> ! {
     // Acquire the GPIOC peripheral
     let mut gpioc = dp.GPIOC.split(&mut rcc.apb2);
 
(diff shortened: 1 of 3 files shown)


TheZoq2:
It looks like our documentation for using `openocd` is incomplete

```
openocd -f interface/$INTERFACE.cfg -f target/stm32f1x.cfg
```

Does not work, at least not without `$INTERFACE` set and i'm not quite sure what it should be set to. I run a custom `blue_pill.cfg` from an older project which *works* but it's probably not the right way to do it. Is the right way to do it documented somewhere? `cortex-m-quickstart` simply  says run `(..)` which is also clearly wrong. Is there a guide for it somewhere?

TeXitoi:
I use that: https://github.com/TeXitoi/blue-pill-quickstart/blob/master/openocd.cfg

Without a config file, using a ST-Link v2 clone, it would be

```
openocd -f interface/stlink-v2.cfg -f target/stm32f1x.cfg
```

TeXitoi:
I think the README.md should present the project, put a quick example, and link to the documentation. No need of openocd configuration or Cargo.toml I think.

therealprof:
@TeXitoi With the bluepill it'll depend on which debug probe you connect to it. Often this will be a crappy STLink clone but it could also be something else entirely (like J-Link, DAPlink, BlackMagicProbe without OpenOCD). From this perspective it would make more sense to add this kind of information to the `embeddonomicon`.

TeXitoi:
@therealprof I agree. But maybe we can explain the most common (I don't know which it is) and point somewhere for other alternatives.

TheZoq2:
I agree, having an easily runable example for common hardware would be nice. My guess is that the bluepilll + some stlink V2 is the most common hardware, but I could be completely wrong about that.

PR #39, opened 2019-03-07
Improve README

TheZoq2:
Followup to #37 which improves the README.

- Add badges for docs.rs and crates.io
- Add example code
- Some wording and stylistic changes
- Prevent line wrap in docs code sample

I also moved the example usage instructions from lib.rs to README because I feel like they are more relevant here. I go to docs.rs when i've added the crate as a dependency and want to use it and to the repo when I first encounter the crate which is when i'd want to run the examples.

I kind of feel like the blinky example is a bit out of place, but I also like having some sort of code sample in the README...

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,12 +29,13 @@ $ rustup target add thumbv7m-none-eabi
 # on another terminal
 $ openocd -f interface/$INTERFACE.cfg -f target/stm32f1x.cfg
 
-# flash and debug the "Hello, world" example
-$ cargo run --features $MICROCONTROLLER --example led
+# flash and debug the "Hello, world" example. Change stm32f103 to match your hardware
+$ cargo run --features stm32f103 --example hello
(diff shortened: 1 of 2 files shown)


TheZoq2:
Yes, a release would be nice, did you end up giving me publish access on crates, if so I can take care of it

therealprof:
I thought so, I vaguely remember looking up how that works and it was surprisingly straight forward. ;)

... was probably some other crate, though. You should have received an invitation about now.

TheZoq2:
Yep, looks like crates doesn't send any sort of notification unless you log in and click owner invites, thanks

TheZoq2:
v0.2.1 has been released :tada: Hopefully I did everything correctly :)

therealprof:
@TheZoq2 Looking good! Thanks.

PR #40, opened 2019-03-08
Fix warnings

TheZoq2:
Another small PR before releasing to crates, hopefully you don't mind the spam but code review tends to be useful.

I commented out the things which are clearly DMA related and removed the rest of the unused code.

I'm a bit confused as to why the `mut` aren't required in `i2c.rs` but if the compiler is fine with it, I suppose it's not an issue.

### DIFF
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -108,6 +108,7 @@ pub struct R;
 /// Write transfer
 pub struct W;
 
+/*
 macro_rules! dma {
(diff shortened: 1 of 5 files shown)


therealprof:
> I'm a bit confused as to why the mut aren't required in i2c.rs but if the compiler is fine with it, I suppose it's not an issue.

Neither `byte` nor `first_bytes` are mutated themselves. As long as `byte` implements `DerefMut` the compiler is happy mutating whatever *byte mutably derefs to.

PR #45, opened 2019-03-17
Enable PWM for stm32f100

geomatsi:
Crate stm32f1 handles CCRx registers differently for stm32f103
and for all the other supported MCU families. As a result,
naming conventions to access CCRx fields  are not the same.

Enable support for PWM on stm32f100 using 'feature' switch to
distinguish between stm32f103 and other MCUs at compile time
and use the correct CCRx field names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Changed
+
+- enable PWM on stm32f100
+
 ## [v0.2.1] - 2019-03-08
 
 ### Added
@@ -15,7 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Add support for real time clock
 - Add patches for using STM32F100 chip (PWM disabled)
 
-### Change
+### Changed
 
 - Improve documentation
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -113,7 +113,6 @@ pub mod flash;
 pub mod gpio;
 pub mod i2c;
 pub mod prelude;
-#[cfg(not(feature = "stm32f100"))]
 pub mod pwm;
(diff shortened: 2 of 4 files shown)


therealprof:
Thanks for your contribution. I've one question and one request to make:
> Crate stm32f1 handles CCRx registers differently for stm32f103 and for all the other supported MCU families.

Does this mean stm32f103 is different to all other chips (which essentially matches your implementation), or is stm32f100 different to all other chips (in which case it would be better to match for the stm32f100 feature)?

Could you also please add an entry to CHANGELOG.md?

geomatsi:
Hi @therealprof 

1. In fact, I don't think there is a difference. According to datasheets, general-purpose TIM registers are the same. It looks like it is all about TIM register naming conventions. For instance, for stm32f103 it is [tim2::ccr::CCR](https://docs.rs/stm32f1/0.6.0/stm32f1/stm32f103/tim2/ccr/index.html) and for stm32f100 (and all the other chip families in f1 crate) it is [tim2::ccrX::CCRX](https://docs.rs/stm32f1/0.6.0/stm32f1/stm32f100/tim2/ccr1/index.html). So far I couldn't figure out why it is different by looking at SVD files.

Let me know if it makes sense to fix on stm32f1 level rather then in stm32f1xx-hal.

2. Changelog: will do and resubmit tomorrow. Unless you suggest to follow stm32f1 fixing path...

Regards,
Sergey

therealprof:
@geomatsi Yes, it should eventually be fixed in `stm32-rs`, working around it should be fine for now, please add a comment to your change about this.

burrbull:
https://github.com/stm32-rs/stm32-rs/pull/169
https://github.com/stm32-rs/stm32-rs/pull/167

geomatsi:
Hi @therealprof and all,

Updated Changelog.md file and force-pushed PR.

Regards,
Sergey

PR #47, opened 2019-03-18
[WIP] feature gates

burrbull:
Develop this idea:
https://github.com/rust-embedded/svd2rust/pull/254

`i2c1 = ["i2c"]` should be `i2c1 = ["i2c", "stm32f1/i2c1"]` in fact.

What are your thoughts about this?


TheZoq2:
I'm not really a fan of this, mostly for the reasons @therealprof discussed in the svd2rust issue.

This seems like a lot of added complexity (both in terms of extra code and when using the library). The only real benefit I see is that it reduces the compile times, but an end user would only compile this crate once when adding it as a dependency. I wouldn't be surprised if it increases the compile time over all because you will have to recompile every time a feature is changed.

I may be wrong though, let me know if I'm missing anything

therealprof:
There's one detail I really like about the proposal: Just importing the `port` (instead of all possible individual pins) and using `port`::`pin`. I think we should definitely do that to clean up the feature gated imports.

burrbull:
> I'm not really a fan of this, mostly for the reasons @therealprof discussed in the svd2rust issue.
> 
> This seems like a lot of added complexity (both in terms of extra code and when using the library). The only real benefit I see is that it reduces the compile times, but an end user would only compile this crate once when adding it as a dependency. I wouldn't be surprised if it increases the compile time over all because you will have to recompile every time a feature is changed.
> 
> I may be wrong though, let me know if I'm missing anything

You are almost right. I've started making this PR just for fun after I had generated [this pac](https://github.com/burrbull/stm32f1-gates). I created it because I was forced to compile stm32f1 (blue-pill) on Orange Pi PC2 and I did not have enough RAM to do this.

burrbull:
> There's one detail I really like about the proposal: Just importing the `port` (instead of all possible individual pins) and using `port`::`pin`. I think we should definitely do that to clean up the feature gated imports.

Don't quite see what exactly is in question.
This #48 PR led me to the idea that for the import of pins it is better to use such feature gates as lqfp-100 instead of port. I should check it.

therealprof:
> Don't quite see what exactly is in question.
This #48 PR led me to the idea that for the import of pins it is better to use such feature gates as lqfp-100 instead of port. I should check it.

I kind of liked this change:
```
-        PA0<Alternate<PushPull>>,
-        PA1<Alternate<PushPull>>,
-        PA2<Alternate<PushPull>>,
-        PA3<Alternate<PushPull>>,
+        gpioa::PA0<Alternate<PushPull>>,
+        gpioa::PA1<Alternate<PushPull>>,
+        gpioa::PA2<Alternate<PushPull>>,
+        gpioa::PA3<Alternate<PushPull>>,
```
because it gets rid of the feature gated imports.

Apart from that, not a fan of this change at all since it just makes it a lot harder to use (unless you opt in to everything in which case it is much ado about nothing).

burrbull:
I understand. In any case this PR is only experiment and is meaningless until this feature will be present in svd2rust.

therealprof:
@burrbull I don't think svd2rust support is going to happen: https://github.com/rust-embedded/svd2rust/pull/236

TheZoq2:
I'll go ahead and close this for now, feel free to re-open it if something changes

PR #48, opened 2019-03-18
RFC: add gpio pins for LQFP-100 package

geomatsi:
Add full range of GPIO pins corresponding to LQFP-100 package
including PC/PD/PE. Unify pin definitions for both supported
stm32 mcu families.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Add gpio pins corresponding to LQFP-100 package
 - Implement `core::fmt::Write` for `serial::Tx`
 
 ### Changed
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -1,8 +1,5 @@
 //! # General Purpose I/Os
 
-// TODO the pins here currently correspond to the LQFP-48 package. There should be Cargo features
-// that let you select different microcontroller packages
-
 use core::marker::PhantomData;
 
 use crate::rcc::APB2;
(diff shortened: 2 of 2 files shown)


TheZoq2:
Thanks for the PR!

This sounds like a good idea. As the comment says, it would be nice to have feature gates that specify  the desired package size and maybe even make one of them a default feature. I'm not sure which package we'd want as a default in that case though

therealprof:
I'm not too happy about the feature gate idea, they should just always be there. The fact that they're not routed to accessible pins on some package types is nothing we should be concerned about. IOW: For me this is fine exactly as contributed.

TheZoq2:
Yea, you are probably right now that I think about it some more. Feature gates can be kind of messy and the tiny bit of extra safety you get from them is probably not worth it.

@geomatsi Would you mind removing the todo comment and fixing the conflict in the changelog?

therealprof:
@TheZoq2 We've discussed this various times in other crates. My view here is still: The packages only affect the bonding, the registers are still there (and the pins even usable in many cases) so adding feature gates doesn't improve the safety at all; we really only add feature gates for the known cases (I don't know of any at the moment) where using unavailable features would cause the MCU to misbehave.

TheZoq2:
> so adding feature gates doesn't improve the safety at all

Apart from when you try to use a pin that isn't present in your package. Maybe not *safety*, but it's still an issue that *could* be fixed.

It's a tiny issue though, and I agree that adding a feature gate here is a bad idea.

therealprof:
> Apart from when you try to use a pin that isn't present in your package

It's the same effect as if you forgot to route the pin to whatever component you want to use it with, it'll just dangle in the breeze.

geomatsi:
Hi all,

I played a bit with github 'in-place conflict resolution' feature. Currently PR does not look correct, so please ignore latest changes. I will resubmit after removing TODO from gpio module and rebasing this PR on top of master.

Regards,
Sergey

geomatsi:
Hi @therealprof and all,

PR has been rebased on top of master. Besides, as per suggestion by @TheZoq2 , TODO comment has been removed from src/gpio.rs module.

Regards,
Sergey

TheZoq2:
Looks good to merge, thanks!

PR #49, opened 2019-03-19
feat: fmt::Write for serial Tx

### LINKED ISSUE #18: Feature: fmt::Write for serial::Tx
Hi,
I'm new to this HAL, rather new to Rust and I still have a lot of confusion, so apologize if this does not belong here or is a stupid question. 

I'm using the usart (like in the example) which provides the function `Tx.write(byte)`. But there is no functionality to send a (formatted) string provided.

This can be implemented with something like:
```
use core::fmt::{self, Write};

struct Format_Tx {
    Tx :  stm32f1xx_hal::serial::Tx<USART1>,
}

impl fmt::Write for Format_Tx {
    fn write_str(&mut self, s: &str) -> fmt::Result {
        for byte in s.bytes() {
            block!(self.Tx.write(byte)).ok();
        }
        Ok(())
    }
}
```

This enables the write macro, if you create a `Format_Tx` "relay" struct. For example `write!(my_format_tx, "Formatted {} String", 123);`

Could  `stm32f1xx_hal::serial::Tx`  implement this?
Then writing strings to the serial out would be as easy as `write!(mytx, "foobar")`. I guess this is a common use case.

Regards Jounathaen

chocol4te:
Closes #18, all credit to @therealprof.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- Implement `core::fmt::Write` for `serial::Tx`
+
 ### Changed
 
 - enable PWM on stm32f100
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -42,6 +42,7 @@ use core::ptr;
 use nb;
 use crate::pac::{USART1, USART2, USART3};
 use void::Void;
+use embedded_hal::serial::Write;
 
 use crate::afio::MAPR;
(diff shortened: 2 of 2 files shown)


chocol4te:
Done.

TheZoq2:
Thanks, though it looks like you added it to 0.2.1 which has already been released. It should be under the `unreleased` section instead :)

chocol4te:
🤦‍♂️ My bad, fixed

therealprof:
@chocol4te There's still a merge conflict.

chocol4te:
Rebased :)

PR #51, opened 2019-03-25
RFC: add support for ADC

geomatsi:
Summary
-------------

This PR adds the following bits:
- implement ADC clock configuration in rcc module
- implement ADC support, including multiple ADC blocks
- add ADC example

This PR is based on #19 by @TheZoq2 . As per request from @therealprof , it has been reworked following stm32f0x ADC implementation. Besides, some implementation details follow stm32f4x ADC implementation.

Limitations
-------------
1. Basic one-shot support only: similar to what we have in stm32f0x, but not in stm32f4xx
2. No dual ADC mode support when 2 ADC blocks are available
3. No fool proof against using the same channel in two ADC blocks.

Implementation notes
-------------
1. Split of ADC implementation to simplify review
To simplify review, ADC implementation is split into the following two parts: 
   - 2nd commit: implement one-shot support for ADC1 only
   - 3rd commit: use rust macros to support multiple ADC blocks (including gpio channel assignment) w/o functional changes to ADC handling code
2. Hardware support
Current implementation enables the following ADC blocks:
   - the only available ADC1 for both stm32f100 and stm32f101
   - both ADC1 and ADC2 available on medium-density stm32f103 chips
3. Support for a more capable stm32f10x chips
Note that XL-density stm32f103 chips support ADC3 as well. Besides, gpio-to-channel mapping is different on those chips. Meanwhile it can be also embraced by suggested implementation. For instance, something like the following should work: 
```
#[cfg(feature = "stm32f103XL")]
adc_pins!(ADC1,
    gpioa::PA0<Analog> => 0_u8,
    gpioa::PA1<Analog> => 1_u8,
    gpioa::PA2<Analog> => 2_u8,
    gpioa::PA3<Analog> => 3_u8,
    gpioa::PA4<Analog> => 4_u8,
    gpioa::PA5<Analog> => 5_u8,
    gpioa::PA6<Analog> => 6_u8,
    gpioa::PA7<Analog> => 7_u8,
    gpiob::PB0<Analog> => 8_u8,
    gpiob::PB1<Analog> => 9_u8,
    gpioc::PC0<Analog> => 10_u8,
    gpioc::PC1<Analog> => 11_u8,
    gpioc::PC2<Analog> => 12_u8,
    gpioc::PC3<Analog> => 13_u8,
    gpioc::PC4<Analog> => 14_u8,
    gpioc::PC5<Analog> => 15_u8,
);

#[cfg(feature = "stm32f103XL")]
adc_pins!(ADC3,
    gpioa::PA0<Analog> => 0_u8,
    gpioa::PA1<Analog> => 1_u8,
    gpioa::PA2<Analog> => 2_u8,
    gpioa::PA3<Analog> => 3_u8,
    gpioa::PA4<Analog> => 4_u8,
    gpioa::PA5<Analog> => 5_u8,
    gpioa::PA6<Analog> => 6_u8,
    gpioa::PA7<Analog> => 7_u8,
    gpiob::PB0<Analog> => 8_u8,
    gpiob::PB1<Analog> => 9_u8,
    gpioc::PC0<Analog> => 10_u8,
    gpioc::PC1<Analog> => 11_u8,
    gpioc::PC2<Analog> => 12_u8,
    gpioc::PC3<Analog> => 13_u8,
    gpioc::PC4<Analog> => 14_u8,
    gpioc::PC5<Analog> => 15_u8,
);

#[cfg(feature = "stm32f103XL")]
adc_pins!(ADC3,
    gpioa::PA0<Analog> => 0_u8,
    gpioa::PA1<Analog> => 1_u8,
    gpioa::PA2<Analog> => 2_u8,
    gpioa::PA3<Analog> => 3_u8,
    gpiof::PF6<Analog> => 4_u8,
    gpiof::PF7<Analog> => 5_u8,
    gpiof::PF8<Analog> => 6_u8,
    gpiof::PF9<Analog> => 7_u8,
    gpiof::PF10<Analog> => 8_u8,
    gpioc::PC0<Analog> => 10_u8,
    gpioc::PC1<Analog> => 11_u8,
    gpioc::PC2<Analog> => 12_u8,
    gpioc::PC3<Analog> => 13_u8,
);

#[cfg(feature = "stm32f103XL")]
adc_hal! {
    ADC1: (
        adc1,
        adc1en,
        adc1rst
    ),
    ADC2: (
        adc2,
        adc2en,
        adc2rst
    ),
    ADC3: (
        adc3,
        adc3en,
        adc3rst
    ),
}
```

### DIFF
--- a/examples/adc.rs
+++ b/examples/adc.rs
@@ -1,0 +1,76 @@
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+#[macro_use]
+extern crate cortex_m_rt as rt;
+extern crate cortex_m;
+extern crate cortex_m_semihosting;
+extern crate embedded_hal;
+extern crate panic_semihosting;
+extern crate stm32f1xx_hal;
+
+use core::fmt::Write;
+use cortex_m_semihosting::hio;
+use stm32f1xx_hal::prelude::*;
+
+use rt::ExceptionFrame;
+use stm32f1xx_hal::adc;
+
+#[entry]
+fn main() -> ! {
+    // Aquire peripherals
+    let p = stm32f1xx_hal::stm32::Peripherals::take().unwrap();
+    let mut flash = p.FLASH.constrain();
+    let mut rcc = p.RCC.constrain();
+
+    // Configure ADC clocks
+    // Default value is the slowest possible ADC clock: PCLK2 / 8. Meanwhile ADC
+    // clock is configurable. So its frequency may be tweaked to meet certain
+    // practical needs. User specified value is be approximated using supported
+    // prescaler values 2/4/6/8.
+    let clocks = rcc.cfgr.adcclk(2.mhz()).freeze(&mut flash.acr);
+    hio::hstdout()
+        .map(|mut hio| writeln!(hio, "adc freq: {}", clocks.adcclk().0).unwrap())
+        .unwrap();
+
+    // Setup ADC
+    let mut adc1 = adc::Adc::adc1(p.ADC1, &mut rcc.apb2);
+
+    #[cfg(feature = "stm32f103")]
+    let mut adc2 = adc::Adc::adc2(p.ADC2, &mut rcc.apb2);
+
+    // Setup GPIOB
+    let mut gpiob = p.GPIOB.split(&mut rcc.apb2);
+
+    // Configure pb0, pb1 as an analog input
+    let mut ch0 = gpiob.pb0.into_analog(&mut gpiob.crl);
+
+    #[cfg(feature = "stm32f103")]
+    let mut ch1 = gpiob.pb1.into_analog(&mut gpiob.crl);
+
+    loop {
+        let data: u16 = adc1.read(&mut ch0).unwrap();
+        hio::hstdout()
+            .map(|mut hio| writeln!(hio, "adc1: {}", data).unwrap())
+            .unwrap();
+
+        #[cfg(feature = "stm32f103")]
+        {
+            let data1: u16 = adc2.read(&mut ch1).unwrap();
+            hio::hstdout()
+                .map(|mut hio| writeln!(hio, "adc2: {}", data1).unwrap())
+                .unwrap();
+        }
+    }
+}
+
+#[exception]
+fn HardFault(ef: &ExceptionFrame) -> ! {
+    panic!("{:#?}", ef);
+}
+
+#[exception]
+fn DefaultHandler(irqn: i16) {
+    panic!("Unhandled exception (IRQn = {})", irqn);
+}
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -4,14 +4,17 @@ use embedded_hal::adc::{Channel, OneShot};
 
 use crate::gpio::Analog;
 use crate::gpio::{gpioa, gpiob, gpioc};
-
 use crate::rcc::APB2;
 
 use crate::stm32::ADC1;
+#[cfg(any(
+    feature = "stm32f103",
+))]
+use crate::stm32::ADC2;
 
-/// Analog to Digital converter interface
-pub struct Adc {
-    rb: ADC1,
+/// ADC configuration
+pub struct Adc<ADC> {
+    rb: ADC,
     sample_time: AdcSampleTime,
     align: AdcAlign,
 }
@@ -41,7 +44,7 @@ pub enum AdcSampleTime {
 }
 
 impl AdcSampleTime {
-    /// Get the default sample time (currently 239.5 cycles)
+    /// Get the default sample time (currently 28.5 cycles)
     pub fn default() -> Self {
         AdcSampleTime::T_28
     }
@@ -63,16 +66,16 @@ impl From<AdcSampleTime> for u8 {
 }
 
 #[derive(Clone, Copy, Debug, PartialEq)]
-/// ADC Result Alignment
+/// ADC data register alignment
 pub enum AdcAlign {
-    /// Left aligned results (most significant bits)
-    Left,
-    /// Right aligned results (least significant bits)
+    /// Right alignment of output data
     Right,
+    /// Left alignment of output data
+    Left,
 }
 
 impl AdcAlign {
-    /// Get the default alignment (currently right aligned)
+    /// Default: right alignment
     pub fn default() -> Self {
         AdcAlign::Right
     }
@@ -81,8 +84,8 @@ impl AdcAlign {
 impl From<AdcAlign> for u8 {
     fn from(val: AdcAlign) -> Self {
         match val {
-            AdcAlign::Left => 1,
             AdcAlign::Right => 0,
+            AdcAlign::Left => 1,
         }
     }
 }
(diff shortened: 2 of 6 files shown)


src/adc.rs
```diff
@@ -0,0 +1,407 @@
...
+            impl Channel<$ADC> for $pin {
+                type ID = u8;
+
+                fn channel() -> u8 { $chan }
+            }
+        )+
+    };
+}
+
+#[cfg(any(
```
TheZoq2: Since ADC1 is available on all pins, perhaps we can skip this

src/adc.rs
```diff
@@ -0,0 +1,407 @@
...
+            }
+
+            impl<WORD, PIN> OneShot<$ADC, WORD, PIN> for Adc<$ADC>
+            where
+                WORD: From<u16>,
+                PIN: Channel<$ADC, ID = u8>,
+                {
+                    type Error = ();
+
+                    fn read(&mut self, _pin: &mut PIN) -> nb::Result<WORD, Self::Error> {
```
TheZoq2: If I am reading this correctly, this function is blocking despite returning an `nb::Result` which seems weird to me.

src/rcc.rs
```diff
@@ -414,6 +441,11 @@ impl Clocks {
         self.sysclk
     }
 
+    /// Returns the system (core) frequency
```
TheZoq2: Incorrect comment

src/lib.rs
```diff
@@ -92,6 +92,7 @@
 //! [examples]: https://github.com/stm32-rs/stm32f1xx-hal/tree/master/examples
 
 #![no_std]
+#![allow(non_camel_case_types)]
```
TheZoq2: Why do we need this? it seems like sticking to standard rust formatting would be desireable

examples/adc.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[entry]
+fn main() -> ! {
+    // aquire peripherals
+    let p = stm32f1xx_hal::stm32::Peripherals::take().unwrap();
+    let mut flash = p.FLASH.constrain();
+    let mut rcc = p.RCC.constrain();
+
+    // configure ADC clocks
+    let clocks = rcc.cfgr.adcclk(2.mhz()).freeze(&mut flash.acr);
```
TheZoq2: I don't think setting the ADC clock in an example is a good idea unless it's required for the operation of the ADC.

src/adc.rs
geomatsi: Ok. Will do.

src/adc.rs
geomatsi: Well, this is how _read_ is declared in embedded-hal OneShot trait. This nb::Result is not necessarily about blocking or non-blocking function. This is just a minimal extension of std::io::Result which adds WouldBlock error code, see [nb docs](https://docs.rs/nb/0.1.1/nb/).

src/rcc.rs
geomatsi: Ooops. Good catch! Will fix.

src/lib.rs
geomatsi: I have been using the same AdcSampleTime enum as in stm32f0xx-hal. So this non_camel_case_types is needed to make rust happy with T_n names. I could drop it and use other names for AdcSampleTime enum, e.g. Tn. But it looks like using non-camel case types is quite common for stm32fxxx HAL implementations: see both stm32f0xx and stm32f4xx.

examples/adc.rs
geomatsi: Why not ? Default value is the slowest possible ADC clock (pclk2 / 8), just to be on the safe side and to ensure that 14MHz threshold is not exceeded. But ADC clock is configurable, so its frequency may be tweaked to meet certain practical needs, e.g. speed vs power savings choice.

geomatsi:
Thanks for review, I will address comments and re-submit. Let me know if you are OK with multiple force-push updates for the same PR.

therealprof:
@geomatsi Feel free to force-push away.

examples/adc.rs
TheZoq2: Alright, sounds reasonable. Though it would be nice to have a comment that explains that in the exammple

src/adc.rs
TheZoq2: Maybe I'm misinterpreting things then, but I as a user would assume that a function returning `nb::Result` is actually non-blocking.

I could see someone implementing a device driver using the `embedded_hal` traits that relies on the ADCs being non blocking for doing something like concurrent reads.

src/lib.rs
TheZoq2: I think I would prefer using `Tn`, but I'd be fine with `T_n`, especially if it is done that way in more crates. However, in that case I would prefer to have the `[allow]` directive on the type rather than as a global thing.

geomatsi:
Hi all,

This PR has been updated. Here is a brief summary of changes.
v1 -> v2: introduce changes according to @TheZoq2  comments
- ADC1 is available on all the pins, so remove cfg(any(..)) for it
- fix comment in rcc.rs: adc frequency, not core frequency
- use allow(non_camel_case_types) only for AdcSampleTime enum, do not make it global
- add detailed comment to adc example: why and how to configure adc clocks
- update adc clocks commit message: why and how to configure adc clocks

Regards,
Sergey

src/adc.rs
geomatsi: Hi @TheZoq2, @therealprof , and all,
Ok, lets finally decide on this one. To block, or not to block: that is the question.
* To block until conversion is completed
Apparently nb::Result can be returned from blocking call: in this case we do not return WouldBlock. So far this approach has been taken by other stm32 HALs including stm32f0xx-hal, stm32f4xx-hal, stm32l0xx-hal
* To use completion bits and return WouldBlock if conversion is not ready
This approach has certain advantages, see @TheZoq2 comments. In this case I will adapt the code from #19 and update adc example.

Comments ? Thoughts ?

Regards,
Sergey

therealprof: @geomatsi If you can implement it truly non-blocking, then go for it. I believe the blocking in the other hals code only happens because any other implementation is not race free. The problem here usually is to know whether the call is supposed to start a new sampling or collect the data of a previous sampling... Of course we can always start blocking and improve in later iteration if someone can think of a clever solution.

TheZoq2: I agree, non blocking would be ideal but it does have some pitfalls, for example [this one](https://github.com/japaric/stm32f103xx-hal/pull/114#issuecomment-450547784) from my original PR.

If you can find a nice way to make it non-blocking, I would prefer it, but i'm also fine with the current solution

geomatsi:
Hi @therealprof , @TheZoq2 

Ok, then lets start with blocking version and merge current solution. 

I will experiment with multiple channel readings while working on  [vacuum cleaner firmware](https://github.com/geomatsi/e.ziclean-cube). So I will check out non-blocking implementations including #19 as well as mentioned in your comments.

Regards,
Sergey

TheZoq2:
Looks good to me as well, thanks!

PR #52, opened 2019-03-28
Adds some simple methods to Timer.

thalesfragoso:
Very simple methods but potentially useful, especially the one to clear the update interrupt flag.
I was having trouble using the timer with interrupts, because the Timer takes ownership of the registers and the `tim` field is private.
Sorry if I got something wrong, I am kinda new to rust.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Add feature for using STM32F101 chip
 - Add gpio pins corresponding to LQFP-100 package
 - Implement `core::fmt::Write` for `serial::Tx`
-- Add methods `stop`, `release` and `clear_uif` to `Timer` (`clear_uif` only on `Timer<TIMX>`)
+- Add methods `stop`, `release` and `clear_update_interrupt_flag` to `Timer` (`clear_update_interrupt_flag` does not apply to `Timer<SYST>`)
+- Add timer interrupt example using RTFM
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Add feature for using STM32F101 chip
 - Add gpio pins corresponding to LQFP-100 package
 - Implement `core::fmt::Write` for `serial::Tx`
+- Add methods `stop`, `release` and `clear_uif` to `Timer` (`clear_uif` only on `Timer<TIMX>`)
 
 ### Changed
 
--- a/examples/timer-interrupt-rtfm.rs
+++ b/examples/timer-interrupt-rtfm.rs
@@ -1,0 +1,99 @@
+//! Uses the timer interrupt to blink a led with different frequencies.
+//!
+//! This assumes that a LED is connected to pc13 as is the case on the blue pill board.
+//!
+//! Note: Without additional hardware, PC13 should not be used to drive an LED, see page 5.1.2 of
+//! the reference manaual for an explanation. This is not an issue on the blue pill.
+
+#![no_std]
+#![no_main]
+
+// you can put a breakpoint on `rust_begin_unwind` to catch panics
+extern crate panic_halt;
+
+use cortex_m::asm::wfi;
+use rtfm::app;
+
+use stm32f1xx_hal::{
+    prelude::*,
+    pac,
+    timer::{ Timer, Event },
+    gpio::{ gpioc::PC13, State, Output, PushPull },
+};
+
+#[app(device = stm32f1xx_hal::pac)]
+const APP: () = {
+
+    static mut LED: PC13<Output<PushPull>> = ();
+    static mut TIMER_HANDLER: Timer<pac::TIM1> = ();
+    static mut LED_STATE: bool = false;
+    
+    #[init]
+    fn init() -> init::LateResources {
+
+        // Take ownership over the raw flash and rcc devices and convert them into the corresponding
+        // HAL structs
+        let mut flash = device.FLASH.constrain();
+        let mut rcc = device.RCC.constrain();
+
+        // Freeze the configuration of all the clocks in the system and store the frozen frequencies
+        // in `clocks`
+        let clocks = rcc.cfgr.freeze(&mut flash.acr);
+
+        // Acquire the GPIOC peripheral
+        let mut gpioc = device.GPIOC.split(&mut rcc.apb2);
+
+        // Configure gpio C pin 13 as a push-pull output. The `crh` register is passed to the
+        // function in order to configure the port. For pins 0-7, crl should be passed instead
+        let led = gpioc.pc13.into_push_pull_output_with_state(&mut gpioc.crh, State::High);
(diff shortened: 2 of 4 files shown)


therealprof:
Looks okay to me but there's no need to release `Clocks` since it's a copyable resource.

thalesfragoso:
> Looks okay to me but there's no need to release `Clocks` since it's a copyable resource.

It makes sense, I did not realize that before. I have changed that bit in the new commit.

src/timer.rs
```diff
@@ -72,6 +72,17 @@ impl Timer<SYST> {
             Event::Update => self.tim.disable_interrupt(),
         }
     }
+
+    /// Stops the timer
+    pub fn stop(&mut self) {
+        self.tim.disable_counter();
+    }
+
+    /// Releases the SYST and clocks
```
therealprof:
```suggestion
    /// Releases the SYST
```

src/timer.rs
```diff
@@ -133,6 +144,22 @@ macro_rules! hal {
...
+                }
+
+                /// Clears Update Interrupt Flag
+                pub fn clear_uif(&mut self) {
+                    self.tim.sr.modify(|_, w| w.uif().clear_bit());
+                }
+
+                /// Releases the TIM Peripheral and clocks
+                pub fn release(self) -> $TIMX {
+                    self.tim.cr1.modify(|_, w| w.cen().clear_bit());
```
therealprof: Can't we simply call `stop()` from above?

TheZoq2:
Looks good to me

thalesfragoso:
I updated the Changelog, don't know if I over-explained it, feel free to modify.

TheZoq2:
Looks good to me, though I did think of another improvement which would be nice. Could you replace the already existing calls to `uif.clear_bit()` with the new `clear_uif` function?

Also, I would argue that calling that function `clear_update_interrupt` might be more intuitive than `clear_uif`

thalesfragoso:
> Looks good to me, though I did think of another improvement which would be nice. Could you replace the already existing calls to `uif.clear_bit()` with the new `clear_uif` function?
> 
> Also, I would argue that calling that function `clear_update_interrupt` might be more intuitive than `clear_uif`

How about `clear_update_interrupt_flag` ? Flag is a very common word and would be the first one I would search for in a function to do this procedure.

TheZoq2:
Yea, that is probably even better!

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Add feature for using STM32F101 chip
 - Add gpio pins corresponding to LQFP-100 package
 - Implement `core::fmt::Write` for `serial::Tx`
+- Add methods `stop`, `release` and `clear_uif` to `Timer` (`clear_uif` only on `Timer<TIMX>`)
```
TeXitoi: unsync

src/timer.rs
```diff
@@ -72,6 +72,17 @@ impl Timer<SYST> {
...
     }
+
+    /// Stops the timer
+    pub fn stop(&mut self) {
+        self.tim.disable_counter();
+    }
+
+    /// Releases the SYST
+    pub fn release(mut self) -> SYST {
+        self.tim.disable_counter();
```
TeXitoi: minor: better to `self.stop()` as any bug on stop must be propagated here.

thalesfragoso:
I will modify the code to use the new methods as proposed and I think I will write an example for the Timer interrupt with rtfm and add it to the PR.
I will probably do it over the weekend.

thalesfragoso:
I ended up getting some spare time today, so there you go. Sorry for the bunch of commits, but I guess you gonna squash them all anyways. 
I tested the example on a bluepill and it worked as expected.

TheZoq2:
Thanks, looks mostly good to me!

@therealprof What's our policy on adding dependencies for examples, #2 didn't really come to a conclusion. I believe that rtfm was originally removed because it didn't compile on stable, but that's solved now.

thalesfragoso:
I kinda understand the dependencies for examples thing, but in the present scenario I would think that it is somewhat bad to push newcomers to implement interrupts (with handlers and stuff) without rtfm.

therealprof:
@TheZoq2 I'm generally not too hot on having examples in this crate (apart from in the docs). Ideally we should have BSP crates which go nuts on examples. I do have a a basic https://github.com/therealprof/nucleo-f103rb crate, if people are interested I could make this free for all, or just create a new blue pill crate.

However for now it's okay for me to just add back the RTFM dependency.

therealprof:
@thalesfragoso The problem is that a universal HAL crate is somewhat tricky to use anyway, especially for beginners; so instead of dragging newcomers into examples in here we really should demonstrate how to get properly started with your own projects by having a proper BSP crate.

examples/timer-interrupt-rtfm.rs
```diff
@@ -0,0 +1,99 @@
...
+
+        // Count used to change the timer update frequency
+        static mut COUNT: u8 = 0;
+
+        if *resources.LED_STATE {
+            // Uses resourcers managed by rtfm to turn led off (on bluepill)
+            resources.LED.set_high();
+            *resources.LED_STATE = false;
+        } else {
+            resources.LED.set_low();
```
TeXitoi: You can call `resources.LED.toggle();` and remove LED_STATE

examples/timer-interrupt-rtfm.rs
```diff
@@ -0,0 +1,99 @@
...
+            TIMER_HANDLER: timer,
+        }
+    }
+
+    #[idle]
+    fn idle() -> ! {
+
+        loop {
+            // Waits for interrupt
+            wfi();
```
TeXitoi: You can just not define an `idle()` function.

TheZoq2:
Yea, eventually we should probably take care of this example stuff, and a BSP with examples does sound like a good idea. I do agree that having a RTFM example for interrupts is much better than one without RTFM :)

I guess all that's left to do then is to merge this, thanks for the help! @thalesfragoso

examples/timer-interrupt-rtfm.rs
thalesfragoso: Does PC13<Output<MODE>> have toggle ? I remember looking at the docs when I was writing other code and could not find it.

examples/timer-interrupt-rtfm.rs
thalesfragoso: Does it automatically call wfi() ?

examples/timer-interrupt-rtfm.rs
TeXitoi: I've used it in the past https://github.com/TeXitoi/blue-pill-quickstart/blob/master/src/main.rs#L41

examples/timer-interrupt-rtfm.rs
TeXitoi: https://japaric.github.io/cortex-m-rtfm/book/en/by-example/app.html#idle

> When no idle function is declared, the runtime sets the SLEEPONEXIT bit and then sends the microcontroller to sleep after running init.

PR #55, opened 2019-04-24
Restore Dma

burrbull:


Cargo.toml
```diff
@@ -19,10 +19,11 @@ name = "timer-interrupt-rtfm"
...
 
 [dependencies]
-cortex-m = "0.5.8"
+cortex-m = ">=0.5.8,<0.7"
 nb = "0.1.1"
-cortex-m-rt = "0.6.7"
-stm32f1 = "0.6.0"
+cortex-m-rt = "0.6.8"
+#stm32f1 = "0.7.0"
+stm32f1 = {path = "/home/burrbull/stm32-rs/stm32f1"}
```
eldruin: A local path slipped in here

examples/adc.rs
```diff
@@ -1,26 +1,23 @@
 #![deny(unsafe_code)]
+#![deny(warnings)]
```
eldruin: I discovered not long ago that this is an anti-pattern. See [here](https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md)

Cargo.toml
burrbull: Sorry, this is an accident.

examples/adc.rs
burrbull: Then I will remove this from other examples.

burrbull:
The next three days I will not be able to commit.
Therefore, if minor fixes are needed, please correct them yourself or wait for me.

PR #56, opened 2019-04-26
Fix gpio configuration for pwm input

gbip:
It seems that I made a mistake in #10 as the gpio where not properly configured. I don't know why our tests worked, but this PR fix this issue.

Here is a screenshot of the reference manual, p166 (table 23).

image


therealprof:
Interesting that we missed it as this makes more sense. But it's quite possible that PWM output also works in OD mode despite not being documented.

PR #59, opened 2019-04-28
aliases for serial tx/rx

burrbull:


therealprof:
@TheZoq2 What do you think about this change? I don't have any particular opinion here.

TheZoq2:
I don't really have any strong opinion either.

If i'm reading this right, this would be a breaking change, right? Then maybe it makes sense to wait with merging it until we have a few more breaking changes. That way we have the option of doing a minor release or two in between

burrbull:
This change is not breaking. Only shortcuts for type names to simplify their use in statics.

TheZoq2:
Oh, I missread it then, in that case, i'd be happy to merge it if you add a changelog entry :)

PR #60, opened 2019-04-30
RFC: docs and examples for custom selection of PWM channels

geomatsi:
Hi all,

In practical applications various selections of PWM channels may be needed. For instance, one or two channels attached to gpio pins for non-default remapping. It is possible to keep adding implementations of the 'Pins' trait into pwm module for all the needed channel/pin combinations. However it does not look practical unless there exist some clever way to describe all possible combinations w/o boilerplate code.

Another option is to keep only basic implementations in pwm module, e.g. 'all the channels for default pin remapping'. As for the custom channel selections, it is possible to use newtype pattern to implement 'Pins'  trait in the user application. This PR adds documentation and example demonstrating the use of newtype pattern for creation of custom PWM channel configuration.

Thoughts ? Comments ?

Regards,
Sergey

### DIFF
--- a/src/pwm.rs
+++ b/src/pwm.rs
@@ -2,14 +2,14 @@
   # Pulse width modulation
 
   The general purpose timers (`TIM2`, `TIM3`, and `TIM4`) can be used to output
-  pulse width modulated signals on some pins. The timers support up to 4 simultaneous
-  pwm outputs in separate `Channels`
+  pulse width modulated signals on some pins. The timers support up to 4
+  simultaneous pwm outputs in separate `Channels`
 
   ## Usage for pre-defined channel combinations
 
-  Start by setting all the pins for the timer you want to use to alternate
-  push pull pins. The `Pins` trait shows the mapping between timers, output pins
-  and channels.
+  Crate defines only basic channel combinations for default AFIO remappings,
+  where all the channels are enabled. Start by setting all the pins for the
+  timer you want to use to alternate push pull pins:
 
   ```rust
   let gpioa = ..; // Set up and split GPIOA
@@ -21,7 +21,7 @@
   );
   ```
 
-  Then call the `pwm` function on the corresponding timer
+  Then call the `pwm` function on the corresponding timer:
 
   ```
     let device: pac::Peripherals = ..;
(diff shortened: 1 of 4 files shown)


TheZoq2:
I like the idea of this, however i'm not super familiar with how the pwm module works so I don't think I can give a ton of feedback on it.

Would you be able to add some documentation to the example explaining how to set the `REMAP` and `C<1-4>` values in the newtype?

geomatsi:
Hi @TheZoq2 ,
Good point. I will add a more detailed explanation with references to MCU datasheet.

Regards,
Sergey

therealprof:
@geomatsi Very nice, I like it. Can't test it myself at the moment, though.

geomatsi:
Hi all,

I added a bit more docs and examples in order to clarify how to define custom channel selections and how to implement `Pins` trait for them. Suggested approach has been tested using on stm32f101 hardware, e.g. here is a [single channel example](https://github.com/geomatsi/e.ziclean-cube/blob/master/src/bin/test-hal-pwm-brushes.rs).

By the way, I would appreciate if you could take a look at docs phrasing and check if anything needs correction from the language point of view.

Regards,
Sergey

kinnison:
A brief skim of this looks good from a language perspective.  I can confirm that this clarified things to the point that I was able to implement the Pwm capability I desired for my STM32F103 based firmware.

:+1:

src/pwm.rs
```diff
@@ -126,7 +196,7 @@ impl Pins<TIM4>
         PB9<Alternate<PushPull>>,
     )
 {
-    const REMAP: u8 = 0b0;
+    const REMAP: u8 = 0b00;
```
kinnison: I think this change is wrong.  `TIM4_REMAP` is, afaict, only one bit.

Zero means PB6,7,8,9 ånd One means PD12,13,14,15

kinnison: I mean, the change is harmless enough, but it could lead somene to think that the value was wider than one bit.

geomatsi: Good catch ! I will push a fixup.

src/pwm.rs
```diff
@@ -2,14 +2,14 @@
...
+  pulse width modulated signals on some pins. The timers support up to 4
+  simultaneous pwm outputs in separate `Channels`
 
-  ## Usage
+  ## Usage for pre-defined channel combinations
 
-  Start by setting all the pins for the timer you want to use to alternate
-  push pull pins. The `Pins` trait shows the mapping between timers, output pins
-  and channels.
+  Crate defines only basic channel combinations for default AFIO remappings,
```
TheZoq2: I think this should say "This crate only defines..."

src/pwm.rs
```diff
@@ -31,16 +31,104 @@
...
+    const C1: bool = false;
+    const C2: bool = false;
+    const C3: bool = true;  // use channel C3
+    const C4: bool = true;  // use channel C4
+    type Channels = (Pwm<TIM3, C3>, Pwm<TIM3, C4>);
+  }
+  ```
+
+  REMAP value and channel pins should specified according to the stm32f1xx specification,
+  namely according to the chapter about AFIO and timer alternate function remappings.
```
TheZoq2: It would be nice to have a chapter/section number here

src/pwm.rs
geomatsi: Done

src/pwm.rs
geomatsi: Done: specified AFIO/TIM section from TRM RM0008 Rev 20.

geomatsi:
Hi @TheZoq2 
PR has been updated with fixes according to your comments.

BTW, I am not sure why TravisCI tries to build the same PR changes twice...

Regards,
Sergey

TheZoq2:
Nicely done, thanks!

PR #61, opened 2019-05-02
Reexport cortex-m

gbip:
I had trouble compiling a previously working project where I was importing `cortex-m` directly.

`Cargo.toml` :

```toml
[dependencies]
cortex-m = "0.6"

[dependencies.stm32f1xx-hal]
features = ["rt","stm32f103"]
git = "https://github.com/stm32-rs/stm32f1xx-hal"
```
Produced the following error :
```
warning: Linking globals named 'CORE_PERIPHERALS': symbol multiply defined!

error: failed to load bc of "cortex_m.bmw16o9v-cgu.0": 

error: aborting due to previous error
```

Reexporting cortex-m from this crate and using it directly instead of depending directly on `cortex-m` fixes this issue.
Please tell me if you can reproduce it :smile:


therealprof:
That shouldn't be necessary. Have you updated your Cargo.lock? If it still does not work, a `cargo tree` output might be helpful.

gbip:
You are right, running `cargo update` fixed the issue...
My bad !

PR #63, opened 2019-05-04
Add more configuration options to UART

TheZoq2:
This PR adds the the ability to configure the stop bits and parity bits of the serial device. It is inspired by the implementation in the f4 hal but it does differ a bit from that.

The main difference is that this implementation doesn't allow you to specify word length which I did for 2 reasons:

- Reading/sending 9 bits would require using a u16 instead of u8
- "Word length" includes parity bits

The second point means that a device configured to send 9 bit words with one parity bit actually sends 8 bit words and the parity bit. This seems to go against the way UART protocols are usually defined. To avoid that confusion, I decided to just disable the option of changing word size and assume the user wants to send 8 bits of actual data.

Another difference to the f4 impl is that I don't use a separate config submodule. To me, that feels unnecessary, but feel free to convince me otherwise.

Finally, this works until we add support for USART5 and 6 which don't allow 1.5 and 0.5 stop bits. Should we future proof this design for that?

### DIFF
--- a/examples/serial_config.rs
+++ b/examples/serial_config.rs
@@ -75,13 +75,8 @@ fn main() -> ! {
     // Split the serial struct into a receiving and a transmitting part
     let (mut tx, mut rx) = serial.split();
 
-    let mut timer = Timer::tim2(p.TIM2, 1000.hz(), clocks, &mut rcc.apb1);
-
     let sent = b'U';
     block!(tx.write(sent)).ok();
-    // block!(timer.wait());
-    // timer.start(1000.hz());
-    // block!(timer.wait());
     block!(tx.write(sent)).ok();
 
 
--- a/examples/serial_config.rs
+++ b/examples/serial_config.rs
@@ -15,7 +15,8 @@ use nb::block;
 use stm32f1xx_hal::{
     prelude::*,
     pac,
-    serial::{Config, Serial},
+    serial::{self, Serial},
+    timer::Timer,
 };
 use cortex_m_rt::entry;
 
@@ -63,7 +64,10 @@ fn main() -> ! {
         p.USART3,
         (tx, rx),
         &mut afio.mapr,
-        Config::default().baudrate(9600.bps()),
+        serial::Config::default()
+            .baudrate(9600.bps())
+            .stopbits(serial::StopBits::STOP2)
+            .parity_odd(),
         clocks,
         &mut rcc.apb1,
     );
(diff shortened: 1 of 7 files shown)


examples/serial_config.rs
```diff
@@ -0,0 +1,89 @@
...
+    );
+
+    // Split the serial struct into a receiving and a transmitting part
+    let (mut tx, mut rx) = serial.split();
+
+    let mut timer = Timer::tim2(p.TIM2, 1000.hz(), clocks, &mut rcc.apb1);
+
+    let sent = b'U';
+    block!(tx.write(sent)).ok();
+    // block!(timer.wait());
```
therealprof: What's the purpose of these commented out lines?

therealprof: Also if we're not going to use the timer, why initialise it?

src/serial.rs
```diff
@@ -107,6 +107,67 @@ impl Pins<USART3> for (PB10<Alternate<PushPull>>, PB11<Input<Floating>>) {
...
+
+    pub fn stopbits(mut self, stopbits: StopBits) -> Self {
+        self.stopbits = stopbits;
+        self
+    }
+}
+
+impl Default for Config {
+    fn default() -> Config {
+        let baudrate = 19_200_u32.bps();
```
therealprof: 19200 baud is kind of an odd choice, I haven't seen that in common use since the age of the acoustic coupler. Maybe either use 9600 or 115200 baud instead which are the two common default baud rates used nowadays (with a slight preference for the latter).

examples/serial_config.rs
TheZoq2: Good catch, I'll remove the comments and timer

src/serial.rs
TheZoq2: You're absolutely right, I have no idea what made me chose 19200. 115200 makes sense to me.

TheZoq2:
I implemented the changes you suggested, does everything look OK now?

Also, the changelog is starting to get big, perhaps we should do a release soon. Though maybe we should fix all the deprecation warnings introduced by `embedded_hal` `2.2.3` first

therealprof:
Release sounds good to me. Not sure we really want to deal with the fall out and actually how to deal with it. I tried changing the `stm32f0xx-hal` to `digital::v2` but failed in a compatible manner without breaking all depending crates.

TheZoq2:
Yea, that change seems strange to me, especially since it's a minor change. It was very strange to do a cargo clean and discover hundreds of warnings.

I tried updating it by just adding `Void` return types, but it will, as you say, break all depending crates.

I'll do a release with what we have for now, then we can figure out what to do about this later

TheZoq2:
Looks like you forgot to bump the the version in the README, should we push a 0.3.1 to fix that or just leave it? @therealprof

therealprof:
Huh, I did? Shouldn't matter too much.

PR #67, opened 2019-05-29
adc: add support for channels 16 and 17 in ADC1

geomatsi:
Add reading methods to read from ADC1 special channels:
- channel 16: temperature sensor
- channel 17: internal reference voltage

Note that these methods should be implemented only for ADC1 block
since in ADC2 and ADC3 those channels are connected to Vss.

### DIFF
--- a/examples/adc_temperature.rs
+++ b/examples/adc_temperature.rs
@@ -9,7 +9,7 @@ use stm32f1xx_hal::{
     pac,
     adc,
 };
-use cortex_m_rt::{entry,exception,ExceptionFrame};
+use cortex_m_rt::entry;
 
 use cortex_m_semihosting::hprintln;
 
@@ -44,14 +44,4 @@ fn main() -> ! {
 
         hprintln!("temp: {}", temp).unwrap();
     }
-}
-
-#[exception]
-fn HardFault(ef: &ExceptionFrame) -> ! {
-    panic!("{:#?}", ef);
-}
-
-#[exception]
-fn DefaultHandler(irqn: i16) {
-    panic!("Unhandled exception (IRQn = {})", irqn);
 }
--- a/examples/adc_temperature.rs
+++ b/examples/adc_temperature.rs
@@ -20,18 +20,7 @@ fn main() -> ! {
     let mut flash = p.FLASH.constrain();
     let mut rcc = p.RCC.constrain();
 
-    // Several examples of HSE, PCLK1, ADC clocks
-    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(8.mhz()).pclk1(2.mhz()).adcclk(1.mhz()).freeze(&mut flash.acr);
-    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(16.mhz()).pclk1(4.mhz()).adcclk(2.mhz()).freeze(&mut flash.acr);
-    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(24.mhz()).pclk1(9.mhz()).adcclk(3.mhz()).freeze(&mut flash.acr);
-    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(24.mhz()).pclk1(12.mhz()).adcclk(4.mhz()).freeze(&mut flash.acr);
-    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(24.mhz()).pclk1(12.mhz()).adcclk(6.mhz()).freeze(&mut flash.acr);
-    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(36.mhz()).pclk1(12.mhz()).adcclk(6.mhz()).freeze(&mut flash.acr);
-    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(32.mhz()).pclk1(16.mhz()).adcclk(8.mhz()).freeze(&mut flash.acr);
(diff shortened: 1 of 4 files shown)


TheZoq2:
Thanks for the PR!

The code and docs look very nice, though the docs mention that "The recommended sampling time for the temperature sensor is 17.1 μs" and as far as I can tell, that isn't done here. With that addressed, i'd say this can be merged

geomatsi:
Hi @TheZoq2 

Good point. However current ADC implementation in this crate has no knowledge about its clock settings. So one possible option would be to ask user to set proper sampling time from the application, where ADC clock settings are known.

On the other hand I agree that it is not convenient and it make a lot of sense to configure and then restore sampling time in read_temp function. I tried to use maximum sampling time (239.5 cycles) as a conservative setting that could fit all the ADC frequencies. But it didn't work for some reason. So I will take a further look and push more commits as soon as I figure out the solution.

Regards,
Sergey

therealprof:
@geomatsi The functions can take the current clocks a parameter.

geomatsi:
@therealprof Sure, but I would rather suggest to add current clock parameter into ADC constructor. IIUC we should  instantiate ADC after RCC clock configuration is frozen, so it is possible to pass current clock value only once.

geomatsi:
Hi @TheZoq2 , @therealprof , and all,
PR has been updated: more changes (see review comments) and more examples.  Could you please take a look ?

Regards,
Sergey

TheZoq2:
Looks good to me, I think I prefer having adc not take clocks as that would make this a non-breaking change which would be nice. Or do we anticipate adc requiring clock data in more situations?

therealprof:
Taking clocks looks fine for me.

TheZoq2:
In that case, i'm fine with merging this, unless you have any other concerns @therealprof

examples/adc_temperature.rs
```diff
@@ -0,0 +1,57 @@
...
+    // Read temperature sensor
+    loop {
+        let temp = adc.read_temp();
+
+        hprintln!("temp: {}", temp).unwrap();
+    }
+}
+
+#[exception]
+fn HardFault(ef: &ExceptionFrame) -> ! {
```
TheZoq2: Most examples use panic_halt rather than semihosting, which removes the need for defining these exceptions. Is there a specific need for them here?

examples/adc_temperature.rs
```diff
@@ -0,0 +1,57 @@
...
+    // Several examples of HSE, PCLK1, ADC clocks
+    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(8.mhz()).pclk1(2.mhz()).adcclk(1.mhz()).freeze(&mut flash.acr);
+    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(16.mhz()).pclk1(4.mhz()).adcclk(2.mhz()).freeze(&mut flash.acr);
+    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(24.mhz()).pclk1(9.mhz()).adcclk(3.mhz()).freeze(&mut flash.acr);
+    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(24.mhz()).pclk1(12.mhz()).adcclk(4.mhz()).freeze(&mut flash.acr);
+    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(24.mhz()).pclk1(12.mhz()).adcclk(6.mhz()).freeze(&mut flash.acr);
+    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(36.mhz()).pclk1(12.mhz()).adcclk(6.mhz()).freeze(&mut flash.acr);
+    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(32.mhz()).pclk1(16.mhz()).adcclk(8.mhz()).freeze(&mut flash.acr);
+    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(40.mhz()).pclk1(20.mhz()).adcclk(10.mhz()).freeze(&mut flash.acr);
+    //let clocks = rcc.cfgr.use_hse(8.mhz()).sysclk(48.mhz()).pclk1(24.mhz()).adcclk(12.mhz()).freeze(&mut flash.acr);
```
TheZoq2: Are the comments here useful?

geomatsi: Well, these are just several examples of clock settings that can be done with external oscillator. My main purpose was to demonstrate more or less the whole range of ADC frequencies. Besides, I think the example for use_hse is helpful regardless of adc usage.

examples/adc_temperature.rs
geomatsi: No. I will remove them.

geomatsi: Done

examples/adc_temperature.rs
geomatsi: I removed those extra examples in comments. One use_hse example should be enough.

eldruin: I think it may be useful to document the relationships between the possible frequencies for the different registers in case there are some restrictions to the accepted values. This would clarify what you have to do if you want to set the ADC clock to 12 MHz, for example.

Going a bit further, I could also think it would be nice to document these value relationships in general here somewhere as a shortcut to some paragraph about prescaler config in the datasheet.

TheZoq2: I agree, but i'm not sure if this example is the right way to do this. I'd say it should either be in the documentation or a separate example.

TheZoq2:
Looks good to merge, thanks for yet another great PR :)

PR #69, opened 2019-06-03
Fix cortex-m-quickstart link

fxedel:
The current URL (https://docs.rs/cortex-m-quickstart/0.3.1/cortex_m_quickstart/) leads to an outdated version of cortex-m-quickstart docs that for example includes the incorrect instruction to call `git clone cortex-m-quickstart --vers 0.3.0` (it should be `cargo clone ...`).

Since version 0.3.2, the docs fail to build: https://docs.rs/crate/cortex-m-quickstart/

However, the [crates page](https://crates.io/crates/cortex-m-quickstart) links to another documentation page which seems to be up-to-date: https://rust-embedded.github.io/cortex-m-quickstart/cortex_m_quickstart/

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,7 +23,7 @@
 //! features = ["stm32f103", "rt"]
 //! ```
 //!
-//! [cortex-m-quickstart]: https://docs.rs/cortex-m-quickstart/0.3.1
+//! [cortex-m-quickstart]: https://rust-embedded.github.io/cortex-m-quickstart/cortex_m_quickstart/
 //!
 //! ## Usage example
 //!


therealprof:
`cargo clone` only works if you have the crate `cargo-clone` installed. Also the https://rust-embedded.github.io/cortex-m-quickstart/cortex_m_quickstart/ info is not accurate anymore. In particular I think it shouldn't suggest using a nightly or beta version. We probably fix the `cortex-m-quickstart` crate release a new version pointing to crates.io instead.

TheZoq2:
Looks like the README of the cortex-m-quickstart repo is more up to date, perhaps we should point to that instead?

https://github.com/rust-embedded/cortex-m-quickstart

TheZoq2:
The quickstart link is no longer in our readme, apart from a lingoring reference which I remove in cc66906

PR #75, opened 2019-07-07
Changes DMA traits to take &mut reference

luctius:
This change will make it easier to use the read and write functions due to them not moving the struct. This will cause them to work in the same contexts as read and write do (in the case of serial read and writes).

My main issue was in the context of RTFM, where I would like to use the dma read and write functions in the idle thread and had rx_dma and tx_dma as resources. I could not use them due to not owning them.

This is one of my first PR's so if there is a problem, please say so and I will attempt modify it.

------------------------------------------------------------
Changes CircReadDma, ReadDmam WriteDma
from taking self to &mut self.

Signed-off-by: Cor Peters <prive@corpeters.nl>

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -581,12 +581,12 @@ impl<PIN> AdcDma<PIN> where PIN: Channel<ADC1> {
     }
 }
 
-impl<B, PIN> crate::dma::CircReadDma<B, u16> for AdcDma<PIN>
+impl<'a, B, PIN> crate::dma::CircReadDma<'a, B, u16> for AdcDma<PIN>
 where
     B: AsMut<[u16]>,
     PIN: Channel<ADC1>,
 {
-    fn circ_read(mut self, buffer: &'static mut [B; 2]) -> CircBuffer<B, Self> {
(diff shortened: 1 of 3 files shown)


burrbull:
Problems with DMA are are described here: https://docs.rust-embedded.org/embedonomicon/dma.html

Split off Flash to other PR, please.

luctius:
> Split off Flash to other PR, please.

Apologies, that was not supposed to be there.

thalesfragoso:
> > Split off Flash to other PR, please.
> 
> Apologies, that was not supposed to be there.
> 
> ~Having read through the embedonomicon, could you explain to me the switch from &mut self to mut self in the 'Overlapping Use' section? Wouldn't the 'Transfer' struct obtain exclusive use in either case (which was the reasoning for my patch).~ Nevermind, I understand the issue.
> 
> My problem is that the current situation gives me a 'Cannot move out of borrow' error, since in RTFM the resource is owned by the overlaying rtfm framework. Is there any way to solve that issue without changing the current API?

Maybe you can wrap the `rx_dma`, etc with an `Option` and then make that the resource, in the task you can `take()` it and give it back after the transfer is done.

luctius:
That works, thanks!

I am closing this PR since this is obviously not a solution.
Thank you all for helping me!

Issue #77, opened 2019-07-07
GPIO pin PB5 and PB4 does not work correctly


MauroMombelli:
Hi,

after managing to run the demo blink on led PC13, I wanted to also blink led on PB3, PB4 and PB5.

PC13 and PB5 works fine, PB3 and PB4 don't. I am aware they may be used for some debug stuff but I set debug to false to Cargo.toml. Also if pin would be in use by something else, i would expect compilation error

full code:

```
// std and main are not available for bare metal software
#![deny(unsafe_code)]  //  Don't allow unsafe code in this file.
//#![deny(warnings)] //  If the Rust compiler generates a warning, stop the compilation with an error.
#![no_std]

#![no_main] 

use nb::block;



use stm32f1xx_hal::{

    prelude::*,

    pac,

    timer::Timer,

};


#[cortex_m_rt::entry]

fn main() -> ! {

    // Get access to the core peripherals from the cortex-m crate

    let cp = cortex_m::Peripherals::take().unwrap();

    // Get access to the device specific peripherals from the peripheral access crate

    let dp = pac::Peripherals::take().unwrap();



    // Take ownership over the raw flash and rcc devices and convert them into the corresponding

    // HAL structs

    let mut flash = dp.FLASH.constrain();

    let mut rcc = dp.RCC.constrain();



    // Freeze the configuration of all the clocks in the system and store

    // the frozen frequencies in `clocks`

    let clocks = rcc.cfgr.freeze(&mut flash.acr);



    // Acquire the GPIOC peripheral

    let mut gpiob = dp.GPIOB.split(&mut rcc.apb2);
    let mut gpioc = dp.GPIOC.split(&mut rcc.apb2);



    // Configure gpio C pin 13 as a push-pull output. The `crh` register is passed to the function

    // in order to configure the port. For pins 0-7, crl should be passed instead.

    let mut led = gpioc.pc13.into_push_pull_output(&mut gpioc.crh);
    let mut led1 = gpiob.pb3.into_push_pull_output(&mut gpiob.crl);
    let mut led2 = gpiob.pb4.into_push_pull_output(&mut gpiob.crl);
    let mut led3 = gpiob.pb5.into_push_pull_output(&mut gpiob.crl);
    led1.set_high();
    led2.set_high();
    led3.set_high();
    // Configure the syst timer to trigger an update every second

    let mut timer = Timer::syst(cp.SYST, 1.hz(), clocks);



[...]

PR #79, opened 2019-07-08
Handle debugger pins

TeXitoi:
At startup, pins used by the debugger are in `Debugger` state, prohibiting their
use. `disable_jtag` allow to get PA15, PB3 and PB4 usable as GPIO or alternate
functions at the cost of disabling JTAG.

Discussed in #77 

cc @MauroMombelli

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Breaking changes
 
 - ADC now requires the clock configuration for intialisation
+- `disable_jtag` now transforms PA15, PB3 and PB4 to forbid their use without desactivating JTAG
 
 ### Changed
 
--- a/examples/nojtag.rs
+++ b/examples/nojtag.rs
@@ -17,12 +17,20 @@ fn main() -> ! {
     let p = pac::Peripherals::take().unwrap();
 
     let mut rcc = p.RCC.constrain();
+    let mut gpioa = p.GPIOA.split(&mut rcc.apb2);
     let mut gpiob = p.GPIOB.split(&mut rcc.apb2);
     let mut afio = p.AFIO.constrain(&mut rcc.apb2);
 
-    afio.mapr.disable_jtag();
+    let (pa15, pb3, pb4) = afio.mapr.disable_jtag(gpioa.pa15, gpiob.pb3, gpiob.pb4);
 
-    gpiob.pb4.into_push_pull_output(&mut gpiob.crl).set_low();
+    let mut pa15 = pa15.into_push_pull_output(&mut gpioa.crh);
+    let mut pb3 = pb3.into_push_pull_output(&mut gpiob.crl);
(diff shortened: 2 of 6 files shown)


TeXitoi:
I've tested nojtag on a bluepill with success.

TeXitoi:
It doesn't work :-( This complies:

```rust
#![deny(unsafe_code)]
#![no_main]
#![no_std]

use panic_halt as _;

use stm32f1xx_hal::{
    prelude::*,
    pac,
};
use cortex_m_rt::entry;

#[entry]
fn main() -> ! {
    let p = pac::Peripherals::take().unwrap();

    let mut rcc = p.RCC.constrain();
    let mut gpiob = p.GPIOB.split(&mut rcc.apb2);
    let mut afio = p.AFIO.constrain(&mut rcc.apb2);

    //let (pa15, pb3, pb4) = afio.mapr.disable_jtag(gpioa.pa15, gpiob.pb3, gpiob.pb4);

    let mut pa15 = gpioa.pa15.into_push_pull_output(&mut gpioa.crh);
    let mut pb3 = gpiob.pb3.into_push_pull_output(&mut gpiob.crl);
    let mut pb4 = gpiob.pb4.into_push_pull_output(&mut gpiob.crl);

    loop {
        pa15.toggle();
        pb3.toggle();
        pb4.toggle();
        cortex_m::asm::delay(8_000_000);
    }
}
```

thalesfragoso:
Maybe if you create a marker (empty) trait and implement it for `Input<MODE>`, `Output<MODE>`,  `Analog` and `Alternate<MODE> (maybe I'm missing another) but not for `Debugger`. And then you can add a bound to MODE on line 189.

Sorry for some lacking explanation, I'm on mobile right now.

TeXitoi:
I see the idea, thanks, will try.

TeXitoi:
@thalesfragoso Thanks it's working.

MauroMombelli:
yay! Suggestion: those are not the only tricky pin, we could also completely disable debugging to get other 2 pin (PA13 and PA14), also the same could be applied to the external clock pin.. let me dig in the docs a moment

TeXitoi:
I've blocked PA13 and PA14, freeing them would only be adding a `disable_jtag_sw` taking the 5 pins.

TeXitoi:
> also i am wandering if this should be proposed to the other implementations too.

I don't understand this sentence. Which other implementations?

MauroMombelli:
> I don't understand this sentence. Which other implementations?

other chip implementation like stm32f3xx-hal etc..

therealprof:
Hm, using `MaybeUninit` will bump the MSRV to 1.36, i.e. the latest and greatest. I personally don't have a problem with that but I can imagine people frowning upon that...

TeXitoi:
@therealprof si you want I dont use it? I can, that's just using uninitialized instead.

TeXitoi:
Or i thinks i can remove this uninitialized trick.

TeXitoi:
@therealprof no more `MaybeUninit`

Issue #84, opened 2019-07-15
hclk calculated incorrectly


kyrias:
Reading through [RM0008](https://www.st.com/resource/en/reference_manual/cd00171190.pdf), it seems like `hclk` is being calculated incorrectly [here](https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/rcc.rs#L191-L204).  The calculation works up until division by 16, after that the reference manual skips to 64 whereas this crate continues with 32, so you get the following correspondence:

hpre_bits | Reference manual | Crate
-|-|-
0xxx | SYSCLK not divided | `(1 << (0b0111 - 0b0111)) == 1`
1000 | SYSCLK divided by 2 | `(1 << (hpre_bits - 0b0111)) == 2`
1001 | SYSCLK divided by 4 | `(1 << (hpre_bits - 0b0111)) == 4`
1010 | SYSCLK divided by 8 | `(1 << (hpre_bits - 0b0111)) == 8`
1011 | SYSCLK divided by 16 | `(1 << (hpre_bits - 0b0111)) == 16`
1100 | SYSCLK divided by 64 | `(1 << (hpre_bits - 0b0111)) == 32`
1101 | SYSCLK divided by 128 | `(1 << (hpre_bits - 0b0111)) == 64`
1110 | SYSCLK divided by 256 | `(1 << (hpre_bits - 0b0111)) == 128`
1111 | SYSCLK divided by 512 | `(1 << (hpre_bits - 0b0111)) == 256`

While it ends up setting `hpre` correctly, it calculates the resulting `hclk` incorrectly which is used in the following calculations of `ppre2` and `ppre1`.

kyrias:
What's sort of devious about this is that `ppre1` and `ppre2` are set to the wrong frequencies, but `pclk1` and `pclk2` end up being calculated to the requested frequency values due to using the same kind of incorrect logic as `hclk`, so the software thinks that the clocks were set properly.

It's a bit tricky to work through the logic, so here's an example where it goes wrong:

```
sysclk                       = 72_000_000
wanted_hclk   = sysclk / 128 =    562_500
wanted_pclk1  = hclk / 4     =    140_625
wanted_pclk2  = hclk / 2     =    281_250

sysclk / wanted_hclk = 128  =>  hpre_bits := 0b1101  # Correct, 0b1101 is divide by 128, which sets HCLK to 562_500.
hclk :=     sysclk / (1 << (hpre_bits - 0b0111))
[...]

PR #85, opened 2019-07-16
Fixes hclk miscalculation

thalesfragoso:
Closes https://github.com/stm32-rs/stm32f1xx-hal/issues/84 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Fix hclk miscalculation
+
 ## [v0.3.0] - 2019-04-27
 
 ### Added
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -204,7 +204,11 @@ impl CFGR {
             })
             .unwrap_or(0b0111);
 
-        let hclk = sysclk / (1 << (hpre_bits - 0b0111));
+        let hclk = if hpre_bits >= 0b1100 {
+            sysclk / (1 << (hpre_bits - 0b0110))
+        } else {
+            sysclk / (1 << (hpre_bits - 0b0111))
+        };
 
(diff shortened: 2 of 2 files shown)


kyrias:
This looks correct to me.

Though I would sort of rather use some kind of abstraction on top of `HPREW` instead, so instead of calculating `hpre_bits`, we would make a `HPREW` value and then have some function that converts a `HPREW` to a divisor which is then used to calculate the `hclk`.  Advantage of that would be that it's much easier to write the correct code that way, and make the correct code be obviously correct, than when manually calculating the bit patterns.

This should be possible to do with const functions, which would allow more compile-time validation of the values even, hm.

thalesfragoso:
I don't think const-fn is there yet. I like the idea of improving readability to find mistakes more easily, but I am not sure if creating a function here is the best for now, since it took only 4 lines to fix the issue. Maybe when we get more flexibility in stable const-fn we can make some refactoring.

@TheZoq2, @therealprof, what do you think ? Can we merge this for now or try to improve ?

therealprof:
@thalesfragoso I think this is good for now.

One question out of curiosity, any reason why you didn't move out the common parts from the branches?

like
```
      let hclk = sysclk / (1 << hpre_bits - if hpre_bits >= 0b1100 { 0b0110 } else { 0b0111 });
```

thalesfragoso:
> @thalesfragoso I think this is good for now.
> 
> One question out of curiosity, any reason why you didn't move out the common parts from the branches?
> 
> like
> 
> ```
>       let hclk = sysclk / (1 << hpre_bits - if hpre_bits >= 0b1100 { 0b0110 } else { 0b0111 });
> ```

Not really, I guess I prefer the former aesthetics, but no strong feelings about that, I can change it to the oneliner if you prefer.

therealprof:
@thalesfragoso I don't care, just curious. ;)

Issue #83, opened 2019-07-15
Write support for the Backup Data Register (BKP_DR, DR1, DR2, ...)


mjepronk:
Currently, we can read from the Backup Data Register (for example DR1) using:
```rust
let device = pac::Peripherals::take().unwrap();
let _ = device.BKP.dr1.read().bits();
```
However, this will not be possible after calling `rcc.bkp.constrain`, because this function takes ownership of `device.BKP`:
```rust
let mut pwr = device.PWR;
let mut backup_domain = rcc.bkp.constrain(device.BKP, &mut rcc.apb1, &mut pwr);
let _ = device.BKP.dr1.read().bits(); // ERROR: borrow of moved value: `device.BKP`
```
Writing to the BKP_DR register should normally be only possible after enabling the power and backup interface clocks and setting the DBP bit the Power Control Register (page 81 of the RM0008 reference manual), i.e. calling `rcc.bkp.constrain`.
I'm not really sure how to fix this. I'm not familiar with `svd2rust` or any lower level code. I'm willing to help but I would for sure need some pointers...

therealprof:
@mjepronk The enable access to specific registers you might want to add methods to https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/backup_domain.rs.

mjepronk:
Any ideas on how that should look like? Something like:
```rust
let mut backup_domain = rcc.bkp.constrain(device.BKP, &mut rcc.apb1, &mut pwr);
let _ = backup_domain.dr1.read().bits();
backup_domain.dr1.write(|w| w.bits(42));
```
Or more something higher level? According to RM0008 there are 42 16-bit registers, so with the simple design we would add 42 methods to the `BackupDomain` struct, which might be a bit too much?

TheZoq2:
I think I would prefer something more high level, perhaps `set_reg(id: u8, value: u8)` and a similar `get_reg` function.

therealprof:
I agree with @TheZoq2 either a function to set this block of registers with a function taking an index or potentially we could also think about an accessor yielding access to the whole block of data registers.

TheZoq2:
I think giving the whole block of data registers might be difficult as we can't give out access to `RTCCR`, `CR` and `CSR` without breaking other abstractions

therealprof:
@TheZoq2 No, I just meant the data registers.

burrbull:
I've collected backup data registers into 2 arrays.
[...]

PR #86, opened 2019-07-16
Add support for reading and writing to the Backup Data Register (#83).

mjepronk:
This is my first attempt at fixing issue #83.

### DIFF
--- a/src/backup_domain.rs
+++ b/src/backup_domain.rs
@@ -25,9 +25,7 @@ pub struct BackupDomain {
 
 macro_rules! write_drx {
     ($self:ident, $drx:ident, $idx:expr, $new:expr) => {
-        unsafe {
-            $self._regs.$drx[$idx].write(|w| w.d().bits($new));
-        }
+        $self._regs.$drx[$idx].write(|w| w.d().bits($new));
     };
 }
 
--- a/src/backup_domain.rs
+++ b/src/backup_domain.rs
@@ -49,6 +49,7 @@ impl BackupDomain {
     /// DRx registers: 0 is DR11, up to 31 for DR42. Providing a number above 31
     /// will panic.
     /// NOTE: not available on medium- and low-density devices!
+    #[cfg(feature = "high")]
     pub fn read_data_register_high(&self, register: usize) -> u16 {
         read_drx!(self, bkp_dr, register)
     }
@@ -66,6 +67,7 @@ impl BackupDomain {
     /// DRx registers: 0 is DR11, up to 31 for DR42. Providing a number above 31
     /// will panic.
     /// NOTE: not available on medium- and low-density devices!
+    #[cfg(feature = "high")]
     pub fn write_data_register_high(&self, register: usize, data: u16) {
         write_drx!(self, bkp_dr, register, data)
     }
--- a/src/backup_domain.rs
+++ b/src/backup_domain.rs
@@ -22,3 +22,53 @@ use crate::pac::BKP;
 pub struct BackupDomain {
     pub(crate) _regs: BKP,
 }
+
+macro_rules! write_drx {
+    ($self:ident, $drx:ident, $idx:expr, $new:expr) => {
+        unsafe {
+            $self._regs.$drx[$idx].write(|w| w.d().bits($new));
+        }
+    };
+}
+
+macro_rules! read_drx {
+    ($self:ident, $drx:ident, $idx:expr) => {
+        $self._regs.$drx[$idx].read().d().bits();
+    };
+}
+
+impl BackupDomain {
+    /// Read a 16-bit value from one of the DR1 to DR10 registers part of the
+    /// Backup Data Register. The register argument is a zero based index to the
+    /// DRx registers: 0 is DR1, up to 9 for DR10. Providing a number above 9
+    /// will panic.
+    pub fn read_data_register_low(&self, register: usize) -> u16 {
+        read_drx!(self, dr, register)
+    }
+
+    /// Read a 16-bit value from one of the DR11 to DR42 registers part of the
+    /// Backup Data Register. The register argument is a zero based index to the
+    /// DRx registers: 0 is DR11, up to 31 for DR42. Providing a number above 31
+    /// will panic.
+    /// NOTE: not available on medium- and low-density devices!
+    pub fn read_data_register_high(&self, register: usize) -> u16 {
+        read_drx!(self, bkp_dr, register)
+    }
+
+    /// Write a 16-bit value to one of the DR1 to DR10 registers part of the
+    /// Backup Data Register. The register argument is a zero based index to the
+    /// DRx registers: 0 is DR1, up to 9 for DR10. Providing a number above 9
+    /// will panic.
+    pub fn write_data_register_low(&self, register: usize, data: u16) {
+        write_drx!(self, dr, register, data)
+    }
+
+    /// Write a 16-bit value to one of the DR11 to DR42 registers part of the
+    /// Backup Data Register. The register argument is a zero based index to the
+    /// DRx registers: 0 is DR11, up to 31 for DR42. Providing a number above 31
+    /// will panic.
+    /// NOTE: not available on medium- and low-density devices!
+    pub fn write_data_register_high(&self, register: usize, data: u16) {
+        write_drx!(self, bkp_dr, register, data)
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Update existing ADC example according to ADC API changes
 - Add new ADC example to read ambient temperature using ADC1 CH16
 - Add `listen` and `unlisten` to `serial::Tx` and `serial::Rx`.
-
+- Add methods `read_data_register` and `write_data_register` to
+  `backup_domain::BackupDomain`, which allow read and write access to the Backup
+  Data Register.
 
 ### Breaking changes
 


TheZoq2:
Thanks for the PR, it looks good to me, and most of my comments are "bikeshedding"

I'm not sure how I feel about `panic!` when indexing out of bouds as panics can be a pain to debug on embedded platforms. Since reg 0 doesn't exist, those errors would probably be pretty common as well.
On the other hand, data registers are something that never changes, so it should be fairly simple to catch the panic in a debugger.

Another option might be to ignore the names of the registers and make the indexes 0 based in the API. As far as I know, no other systems in the processor depends on data in these registers so where we put "1" won't make a difference.

Whichever option we go with, it would be nice with a more clear comment saying which range of registers are available.

Since I'm bikeshedding already, I think I would prefer `read_data_register` over `read_bkp_dr`. Most of the functions in the rest of the crate avoids abbreviated names.

Finally, could you add a line to `CHANGELOG.md` saying that this feature has been implemented?

src/backup_domain.rs
```diff
@@ -22,3 +22,143 @@ use crate::pac::BKP;
 pub struct BackupDomain {
     pub(crate) _regs: BKP,
 }
+
+
+macro_rules! write_drx {
+    ($self:ident, $drx:ident, $new:expr) => ({
+        let old: u16 = $self._regs.$drx.read().bits() as u16;
+        if old != $new {
```
therealprof: Why is it required to check whether the value has changed before writing it?

mjepronk:
No problem, bikeshedding is fine.
I'll change the function names to `read_data_register` and `write_data_register`.
I'll add a line to `CHANGELOG.md`.
What do the others think of the `register: u8` argument, zero or one based index?
What do you think of the array based functions?

src/backup_domain.rs
mjepronk: Good question, I'm not sure if it's done by the lower level code or not. I did it because I heard about wear of the EEPROM, don't know if this will make a difference or not.

src/backup_domain.rs
```diff
@@ -22,3 +22,143 @@ use crate::pac::BKP;
...
+            34 => read_drx!(self, bkp_dr34),
+            35 => read_drx!(self, bkp_dr35),
+            36 => read_drx!(self, bkp_dr36),
+            37 => read_drx!(self, bkp_dr37),
+            38 => read_drx!(self, bkp_dr38),
+            39 => read_drx!(self, bkp_dr39),
+            40 => read_drx!(self, bkp_dr40),
+            41 => read_drx!(self, bkp_dr41),
+            42 => read_drx!(self, bkp_dr42),
+            x  => panic!("Backup Data Register dr{} does not exist.", x),
```
therealprof:
```suggestion
            x  => unreachable!(),
```

src/backup_domain.rs
```diff
@@ -22,3 +22,143 @@ use crate::pac::BKP;
...
+            34 => write_drx!(self, bkp_dr34, data),
+            35 => write_drx!(self, bkp_dr35, data),
+            36 => write_drx!(self, bkp_dr36, data),
+            37 => write_drx!(self, bkp_dr37, data),
+            38 => write_drx!(self, bkp_dr38, data),
+            39 => write_drx!(self, bkp_dr39, data),
+            40 => write_drx!(self, bkp_dr40, data),
+            41 => write_drx!(self, bkp_dr41, data),
+            42 => write_drx!(self, bkp_dr42, data),
+            x  => panic!("Backup Data Register dr{} does not exist.", x),
```
therealprof:
```suggestion
            _  => unreachable!(),
```

src/backup_domain.rs
TheZoq2: Correct me if i'm wrong, but I don't think the bkp_drs are implemented using EEPROM since they lose their data between power cycles, they only keep their values while VBAT has power

src/backup_domain.rs
mjepronk: I wonder why `unreachable!` is better than `panic!` with a descriptive error message?

src/backup_domain.rs
therealprof: Yeah, IIRC it's regular RAM which is powered by the battery backup (if necessary). If it was any kind of non-volatile storage that wouldn't work either because you need to erase a block of memory before you could write to it. 😉

src/backup_domain.rs
TheZoq2: I believe you can provide unreachable with a message as well which makes it even more descriptive.

One problem with messages though, is that the string to be written has to be stored somewhere which might use valuable resources

therealprof: Oh, plenty reasons. Formatting is expensive, `unreachable!()` will just throw out a fixed string which can also be reused if used in multiple places. Also it tells the compiler (and the reader) that this situation can not happen under normal circumstances... You're really not expected to see that `panic!()` (and it's message) in embedded implementations, it's just to prevent undefined behaviour.

therealprof:
> What do the others think of the register: u8 argument, zero or one based index?

0 based. I'm just not so sure about the u8, maybe it would be better to make it `usize` for compatibility with general indexing operations.

src/backup_domain.rs
```diff
@@ -22,3 +22,143 @@ use crate::pac::BKP;
...
+        }
+    }
+
+    /// Read an array of 42 elements (one element for every 16-bit DRx register)
+    /// from the Backup Data Register. The first element in the array (index 0)
+    /// is DR1.
+    pub fn read_bkp_dr_array(&self) -> [u16; 42] {
+        let mut data = [0; 42];
+        for i in 0..42 {
```
therealprof: This is not idiomatic as `clippy` will tell you:
```
   = note: #[warn(clippy::needless_range_loop)] on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_range_loop
```

Should be something like:
```
        for (i, d) in data.iter_mut().enumerate() {
            *d = self.read_bkp_dr((i + 1) as u8);
```
instead

burrbull:
```rust
pub fn read_array(&self, buf: &mut [u16], start: usize) {
```
looks more useful then reading of full register block.

src/backup_domain.rs
mjepronk: OK, thanks for the clarification. This is my first embedded project :-) I will remove the read and the comparison.

src/backup_domain.rs
mjepronk: Thanks for the clarification! I will change as suggested.

therealprof:
Hm, just noticed that travis failed for STM32F100 and STM32F101; are the SVD files broken and/or do we need to need to feature gate the backup domain stuff for STM32F103 and higher?

burrbull:
> Hm, just noticed that travis failed for STM32F100 and STM32F101; are the SVD files broken and/or do we need to need to feature gate the backup domain stuff for STM32F103 and higher?

stm32-rs have regression. STM32F103 and other have different types.
I've fixed it in https://github.com/stm32-rs/stm32-rs/pull/256

mjepronk:
I've found that the BKP_DR11 to BKP_DR42 registers do not work (writing a value to them and reading afterwards gives back 0). I already found that they used a different naming convention in the STM32F1 crate. Does anybody know more about this?

burrbull:
> I've found that the BKP_DR11 to BKP_DR42 registers do not work (writing a value to them and reading afterwards gives back 0). I already found that they used a different naming convention in the STM32F1 crate. Does anybody know more about this?

It can be problem of incorrect BKP register address:
изображение

изображение

burrbull:
https://github.com/stm32-rs/stm32-rs/pull/256/commits/58f8163b78edc2d095b239cb21869c5f7defa89a
should fix problem

burrbull:
Related issue in old repo:
https://github.com/japaric/stm32f103xx/issues/18

mjepronk:
@burrbull Thanks, I have tried your [change](https://github.com/stm32-rs/stm32-rs/commit/58f8163b78edc2d095b239cb21869c5f7defa89a) and I saw that you've renamed `bkp_dr11` - `bkp_dr42` to `dr11` - `dr42`, great! But they still don't work. Also, `dr1` does not seem to work either... 

I'll update this PR with my latest code. The code I'm using to test is:
```rust
        let mut x: [u16; 42] = [0; 42];
        for (i, d) in x.iter_mut().enumerate() {
            *d = (i + 1) as u16;
        }
        backup_domain.write_data_register_array(&x, 0);

        let mut buf: [u16; 42] = [0; 42];
        backup_domain.read_data_register_array(&mut buf, 0);
        for (i, d) in buf.iter().enumerate() {
            hprintln!("{}", d);
        }
```

burrbull:
With this change `dr` will be array.
So you don't need macros.
Access with `dr[i]` and `bkp_dr[j]`
where `i = 0..9`  `j = 0..31`

mjepronk:
Aah, I'm sorry for the trouble, thanks. Why do we split into `dr` and `bkp_dr`? Should we reflect this divide in the API we're defining here too?

burrbull:
```rust
    pub fn read_data_register(&self, register: usize) -> u16 {
        match register {
            1..10  => read_drx!(self, dr, register-1),
            11..42 =>  read_drx!(self, bkp_dr, register-11)
             _=> unreachable!()
        }
   }
```

burrbull:
> Aah, I'm sorry for the trouble, thanks. Why do we split into `dr` and `bkp_dr`? Should we reflect this divide in the API we're defining here too?

Because `svd2rust` still not support deriving for registers:
Waiting for:
https://github.com/rust-embedded/svd2rust/pull/256

src/backup_domain.rs
```diff
@@ -22,3 +22,61 @@ use crate::pac::BKP;
...
+    ($self:ident, $drx:ident, $idx:expr, $new:expr) => {{
+        unsafe {
+            $self._regs.$drx[$idx].write(|w| w.bits($new.into()));
+        }
+    }};
+}
+
+macro_rules! read_drx {
+    ($self:ident, $drx:ident, $idx:expr) => {
+        $self._regs.$drx[$idx].read().bits() as u16;
```
burrbull: Better:
`$self._regs.$drx[$idx].read().d().bits()`

src/backup_domain.rs
```diff
@@ -22,3 +22,61 @@ use crate::pac::BKP;
 pub struct BackupDomain {
     pub(crate) _regs: BKP,
 }
+
+macro_rules! write_drx {
+    ($self:ident, $drx:ident, $idx:expr, $new:expr) => {{
+        unsafe {
+            $self._regs.$drx[$idx].write(|w| w.bits($new.into()));
```
burrbull: `$self._regs.$drx[$idx].write(|w| w.d().bits($new));`
And and `unsafe` is not needed.

src/backup_domain.rs
```diff
@@ -22,3 +22,61 @@ use crate::pac::BKP;
...
+            _       => unreachable!(),
+        }
+    }
+
+    /// Write a 16-bit value to the Backup Data Register. The first argument
+    /// denotes the DRx register to write too, starting from zero, i.e. 0 is
+    /// DR1, up to 41 for DR42. Providing a number above 41 will panic.
+    pub fn write_data_register(&self, register: usize, data: u16) {
+        match register {
+            0...9   => write_drx!(self, dr, register, data),
```
burrbull: `0..=9`

`...` was deprecated.

src/backup_domain.rs
mjepronk: Apparently, I still need `unsafe`. I've added the `.d()` part.

src/backup_domain.rs
mjepronk: Fixed.

src/backup_domain.rs
mjepronk: You're right. I've changed it.

src/backup_domain.rs
```diff
@@ -22,3 +22,67 @@ use crate::pac::BKP;
...
+    ($self:ident, $drx:ident, $idx:expr, $new:expr) => {
+        unsafe {
+            $self._regs.$drx[$idx].write(|w| w.d().bits($new));
+        }
+    };
+}
+
+macro_rules! read_drx {
+    ($self:ident, $drx:ident, $idx:expr) => {
+        $self._regs.$drx[$idx].read().d().bits() as u16;
```
burrbull: Unneeded cast `u16 as u16`

mjepronk:
OK, so the following points are open:

- We should decide if we want to prevent people with low- or medium density STM32's of using the registers `DR11`-`DR42` at compile time. Right now, reading always returns 0 and writing is a noop on these devices.
- The PR for the `stm32-rs` repository from burrbull needs to be merged (and released?) before it makes sense to merge this code.
- There is still code in `stm32-rs` repository that requires `unsafe` to write to either `bkp` or `bkp_dr`. See burrbull's review comments.

Let me know if there is still something I can do!

burrbull:
I think we should split each function on low/high variant, where first one will operate on `dr` and second feature gated function will operate on `bkp_dr`.

TheZoq2:
> We should decide if we want to prevent people with low- or medium density STM32's of using the registers DR11-DR42 at compile time. Right now, reading always returns 0 and writing is a noop on these devices.

I'm not a fan of runtime checks, so I'd prefer a feature gate. If we do do runtime checks, i'd prefer some sort of error over returning valid looking data like 0

> I think we should split each function on low/high variant, where first one will operate on dr and second feature gated function will operate on bkp_dr.

I like this solution

mjepronk:
OK, I separated the functions in `_low` and `_high`.

I dropped the array / buffer functions, because I think it would introduce to much functions for just the backup register. But I think this code is very straightforward to implement by the library user. Also they are not always practical (for example if I want to store a u32 in the backup data register, I would have to go from a `u32` to an array of 2 `u16`'s anyway). If you want them back, shout out!

I'm not sure how to do the feature gate:
- low/medium or high/xl density?
- stm32f101, stm32f102, ..? or 
- high-backup-register?

Sorry, for the lots of back- and forward...

src/backup_domain.rs
```diff
@@ -22,3 +22,57 @@ use crate::pac::BKP;
...
+    };
+}
+
+impl BackupDomain {
+    /// Read a 16-bit value from one of the DR1 to DR10 registers part of the
+    /// Backup Data Register. The register argument is a zero based index to the
+    /// DRx registers: 0 is DR1, up to 9 for DR10. Providing a number above 9
+    /// will panic.
+    pub fn read_data_register_low(&self, register: usize) -> u16 {
+        assert!(register < 10);
```
burrbull: Why you add assert here?
`dr[register]` will panic with `index out of bounds`

mjepronk: OK, I'll remove it.

mjepronk:
What do we do with the feature gate?
When will a new version of `stm32f1` be released?
What else is in the way of merging this?

burrbull:
You can add `#[cfg(feature="stm32f107")]`.

@adamgreig  promised make release this weekend.

mjepronk:
That feature does not exist, right? Besides, I see that it depends on model number and on the amount of flash memory:

> Low-density devices are STM32F101xx, STM32F102xx and STM32F103xx
microcontrollers where the Flash memory density ranges between 16 and 32 Kbytes.
Medium-density devices are STM32F101xx, STM32F102xx and STM32F103xx
microcontrollers where the Flash memory density ranges between 64 and 128 Kbytes.
High-density devices are STM32F101xx and STM32F103xx microcontrollers where the
Flash memory density ranges between 256 and 512 Kbytes.
XL-density devices are STM32F101xx and STM32F103xx microcontrollers where the
Flash memory density ranges between 768 Kbytes and 1 Mbyte.
Connectivity line devices are STM32F105xx and STM32F107xx microcontrollers.

The 84-byte data registers are only available in high-density, XL-density and connectivity line devices.

burrbull:
One of the solution is to add feature gate for each type of microcontroller: stm32f1xxxx (stm32f103c8). They can depend on other features like `mem64`.
But not everyone likes this solution.

TheZoq2:
> One of the solution is to add feature gate for each type of microcontroller: stm32f1xxxx (stm32f103c8). They can depend on other features like mem64.
But not everyone likes this solution.

That seems like a pretty reasonable solution to me, but I haven't heard the arguments against it

burrbull:
I can make initial PR with these features.

therealprof:
@TheZoq2 One of the arguments against it is that we'd have to maintain a database of details even most users will not even know or care about. I'm okay with having features for `medium`, `high`, `xl` and and `connectivity` lines with low being implied if none of the higher features are selected.

burrbull:
@mjepronk please, rebase your changes.

mjepronk:
@burrbull I think my repo was already up-to-date, but I've bumped the dependency `stm32f1` to version 0.8 (because your work on that repo are required for this PR).

burrbull:
This repo is already on 0.8 #88  :)

mjepronk:
Doh, sorry. I've merged...

therealprof:
@mjepronk How about a rebase? 😉

mjepronk:
@therealprof @burrbull Done, sorry for the confusion. I didn't do the `push --force` the first time.

src/backup_domain.rs
```diff
@@ -22,3 +22,53 @@ use crate::pac::BKP;
...
+        write_drx!(self, dr, register, data)
+    }
+
+    /// Write a 16-bit value to one of the DR11 to DR42 registers part of the
+    /// Backup Data Register. The register argument is a zero based index to the
+    /// DRx registers: 0 is DR11, up to 31 for DR42. Providing a number above 31
+    /// will panic.
+    /// NOTE: not available on medium- and low-density devices!
+    pub fn write_data_register_high(&self, register: usize, data: u16) {
+        write_drx!(self, dr, register, data)
```
burrbull: write_drx!(self, bkp_dr, register, data)

mjepronk: Sorry, I'm happy that you're more alert than I am!

TheZoq2:
This looks good to me, do we want to wait for #90 before merging?

src/backup_domain.rs
```diff
@@ -22,3 +22,53 @@ use crate::pac::BKP;
 pub struct BackupDomain {
     pub(crate) _regs: BKP,
 }
+
+macro_rules! write_drx {
+    ($self:ident, $drx:ident, $idx:expr, $new:expr) => {
+        unsafe {
```
burrbull: with 0.8 you don't need unsafe here
https://docs.rs/stm32f1/0.8.0/stm32f1/stm32f103/bkp/dr/struct._DW.html

mjepronk: Fixed, thanks!

TheZoq2:
@mjepronk #90 has been merged which should be the last blocker for this, could you add the correct features for this?

mjepronk:
Yes, I will pick this up ASAP.

mjepronk:
Is there anything left to do before we can merge this?

TheZoq2:
Nope, I just missed the new pushed commits, thanks!

Also, I feel like it's time for another release, any objections to that? @therealprof

therealprof:
@TheZoq2 No objections.

TheZoq2:
Cool, version 0.4.0 has been published :tada:

PR #91, opened 2019-07-29
gpio consts

burrbull:
cc @therealprof 

Can this affect on debug builds?

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -212,18 +212,18 @@ macro_rules! gpio {
                         self,
                         cr: &mut $CR,
                     ) -> $PXi<Alternate<PushPull>> {
-                        let offset = (4 * $i) % 32;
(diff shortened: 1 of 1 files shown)


therealprof:
@burrbull I don't think it will. Variables are automatically `const` by default unless they're `mut`, so IIRC to the compiler they're pretty much the same thing.

burrbull:
I see other:

| example    | with consts | without consts |
|------------|:-----------:|---------------:|
| blinky     |    14182    |          14594 |
| led        |     5042    |           5454 |
| pwm_custom |    18538    |          19246 |

therealprof:
Interesting, guess I'll have to revisit this then; maybe in dev mode there's some special handling which removes the explicit assignment, much like `#[inline(always)]` would.

Anyway, I don't object to your PR, especially with those nasty macros being explicit can be a good thing.

PR #92, opened 2019-07-31
Remove unnessecary unsafe blocks

TheZoq2:
As of the latest stm32f1 update, some unsafe blocks are no longer required this removes those. Since this just fixes compiler warnings, I very much doubt this will cause issues, but i'm making a PR for it just to be sure

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -283,12 +283,12 @@ macro_rules! adc_hal {
                 fn setup_oneshot(&mut self) {
                     self.rb.cr2.modify(|_, w| w.cont().clear_bit());
                     self.rb.cr2.modify(|_, w| w.exttrig().set_bit());
-                    self.rb.cr2.modify(|_, w| unsafe { w.extsel().bits(0b111) });
+                    self.rb.cr2.modify(|_, w| w.extsel().bits(0b111));
 
                     self.rb.cr1.modify(|_, w| w.scan().clear_bit());
                     self.rb.cr1.modify(|_, w| w.discen().set_bit());
 
-                    self.rb.sqr1.modify(|_, w| unsafe { w.l().bits(0b0) });
+                    self.rb.sqr1.modify(|_, w| w.l().bits(0b0));
                 }
 
                 fn set_chan_smps(&mut self, chan: u8) {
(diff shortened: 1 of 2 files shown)


burrbull:
Some of `stm32-rs` patches were not applied for stm32f100

TheZoq2:
Oh, I'm glad I made a PR then :P Is them not being applied a feature or will it be fixed soon?

burrbull:
I'll see later what can be done.

More interesting question is what to do with digital v2?

TheZoq2:
Yea, i'm not totally read up on digital v2. Have any of the other hals dealt with that yet?

burrbull:
I've already tried to port this repo on `v2` but in this case old drivers don't work.

Maybe add `#![allow(deprecated)]` temporary?

TheZoq2:
Yea, I think @therealprof also tried porting it at some point but ran into issues. Allowing deprecated might be our best option.

geomatsi:
Hi @TheZoq2 

It turned out that recently stm32f4xx-hal has been updated: see https://github.com/stm32-rs/stm32f4xx-hal/pull/103. 

Maybe now this is time to do the same and break stm32f1xx-hal world as well ? :)
Let me know if it works for you. Then I will port f4 approach to f1 hal.

Regards,
Sergey

burrbull:
@geomatsi Have you tested it with old drivers (v1)?

TheZoq2:
I don't like breaking the world, but I guess if F4, F3 and L0 are doing it, we might as well. Or does @therealprof disagree?

burrbull:
Here:
https://github.com/stm32-rs/stm32f4xx-hal/blob/2dfbe957ce07fdbbb5b3ea02cd67a3520d196855/src/gpio.rs#L161

you could do
```
self.is_set_low().map(|b| !b)
```
to avoid `unwrap`

geomatsi:
Hi @burrbull 

No, I didn't. I don't have any stm32f4 pet project at the moment. There are compat layers in embedded-hal, including both v2 -> v1 and v1 -> v2. But I am not sure if they are going to work seamlessly w/o updating drivers.

therealprof:
@TheZoq2 I try to be neutral on such a decision. This has caused me enough sweat and pain that I'd rather stay of any discussion.

TheZoq2:
I read through the discussions surrounding this again, and from what I can tell, our current `v1` hal is compatible with `v2` drivers and applications without anyone changing anything, with the "minor" issue that we get a bunch of deprecation warnings.

While if we update to `v2`, we would break compatibility with all the `v1` consumers. Since our crate doesn't produce errors from digital things anyway, upgrading to `v2` doesn't seem to provide any benefit.

If the above is correct, i'd say the best approach is to stay on `v1` until more drivers and applications have migrated, at which point we would break less things with our upgrade.

TeXitoi:
I thinks there is a v2 to v1 adapter that is moteur manual, something like into_v1, thus a hal v2 can be used with v1 drivers.

https://docs.rs/embedded-hal/0.2.3/embedded_hal/digital/v1_compat/index.html

burrbull:
How to use this?

TeXitoi:
I never tried and am in hollidays. I'll try to write an exemple without any promise.

geomatsi:
Hi all,

I double-checked how it is done in [bitbang-hal](https://github.com/sajattack/bitbang-hal). It looks like v1/v2 gpio compat layer in embedded-hal indeed works seamlessly:
- embedded_hal::digital::v2 is directly imported in bitbang-hal and errors are handled, e.g. see i2c/spi implementation
- stm32f1xx-hal v1 gpio can be directly used with bitbang-hal w/o any intermediate layer, e.g. see [i2c example on stmf101](https://github.com/geomatsi/e.ziclean-cube/blob/master/examples/hal/example-hal-i2c-acc-test1.rs)

Regards,
Sergey

TeXitoi:
v2 hal with v1 device mockup:

```rust
#![allow(deprecated)]

use embedded_hal::digital::{v1, v1_compat::OldOutputPin, v2};

struct Pin;

impl v2::OutputPin for Pin {
    type Error = ();
    fn set_low(&mut self) -> Result<(), Self::Error> {
        Ok(())
    }
    fn set_high(&mut self) -> Result<(), Self::Error> {
        Ok(())
    }
}

struct Device<P>(P);
impl<P: v1::OutputPin> Device<P> {
    fn new(pin: P) -> Device<P> {
        Device(pin)
    }
}

fn main() {
    let _device: Device<OldOutputPin<_>> = Device::new(Pin.into());
}
```

TheZoq2:
I believe all unsafe warnings have been fixed with the upgrade to the f1 crate

PR #93, opened 2019-08-02
digital v2

burrbull:


src/gpio.rs
```diff
@@ -143,49 +143,52 @@ macro_rules! gpio {
             }
 
             impl<MODE> OutputPin for $PXx<Output<MODE>> {
-                fn set_high(&mut self) {
+                type Error = ();
```
TheZoq2: I think I read somewhere that the prefered error type for a non-failable pin is `!` rather than `()`. Though the never type isn't fully implemented iirc

burrbull:
> Perhaps it would be a good idea to add that to the docs to make things less confusing for new users.

My English not good enough to write documentation.

TheZoq2:
> My English not good enough to write documentation.

No problem. I was thinking something like this

# Interfacing with v1 traits

`embedded-hal` has two versions of the digital traits, `v2` which is used by this crate and `v1` which is deprecated but still used by a lot of drivers. If you want to use such a driver with this crate, you need to convert the digital pins to the `v1` type.

This is done using `embedded-hal::digital::v1_compat::OldOutputPin`. For example:

```rust
let nss = gpioa.pa4.into_push_pull_output(&mut gpioa.crl);
let mut mfrc522 = Mfrc522::new(spi, OldOutputPin::from(nss)).unwrap();
```

burrbull:
Do I need to add this in README or only in docs?

TheZoq2:
Docs should be enough I think

TeXitoi:
Maybe using the void crate for the error instead of ()?

TheZoq2:
Yea, that might not be a bad idea, especially since we already depend on it. 

What are the problems with using `!` here?

TeXitoi:
I think ! is not usable at this position on stable.

burrbull:
Done. Rebased.

TheZoq2:
I think it might be better to put the new docs in the `digital` module since that's probably where people will look when they run into issues

burrbull:
What `digital` module you mean? `gpio` maybe?

TheZoq2:
Yes

Sent from ProtonMail mobile

-------- Original Message --------

burrbull:
Done.

src/gpio.rs
```diff
@@ -437,72 +455,76 @@ macro_rules! gpio {
...
+                        Ok(unsafe { (*$GPIOX::ptr()).odr.read().bits() & (1 << $i) == 0 })
                     }
                 }
 
                 impl<MODE> InputPin for $PXi<Input<MODE>> {
-                    fn is_high(&self) -> bool {
-                        !self.is_low()
+                    type Error = Void;
+                    fn is_high(&self) -> Result<bool, Self::Error> {
+                        Ok(!(self.is_low()?))
```
geomatsi: Is there any reason  why this snippet differs from all the other places ?
Why not the same here:
```
self.is_low().map(|b| !b)
```

burrbull: The reason is I tired.

burrbull: Fixed.

TheZoq2:
It looks like `examples/blinky_rtc.rs` is missing `unwrap`, but other than that, this looks good to merge

burrbull:
Fixed.

TheZoq2:
Perfect, thanks!

PR #97, opened 2019-08-14
as_slice instead of as_ref

burrbull:
cc @therealprof 

Use as_slice/as_mut_slice for dma buffers, as as_ref restricted by 32-sized buffers.


burrbull:
Done.

burrbull:
Merge, please.

PR #100, opened 2019-08-17
 replace Void with Infallible + small fixes

burrbull:
r? @therealprof

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Breaking changes
 
+- `void::Void` replaced with `Infallible` where it is possible
 - Change timer/pwm init API
 - Remove `set_low` and `set_high` for pins in Alternate output mode
 - Renames `set_seconds` and `seconds` methods on RTC to `set_time` and `current_time`, respectively
--- a/README.md
+++ b/README.md
@@ -110,14 +110,15 @@ fn main() -> ! {
     // in order to configure the port. For pins 0-7, crl should be passed instead.
     let mut led = gpioc.pc13.into_push_pull_output(&mut gpioc.crh);
     // Configure the syst timer to trigger an update every second
-    let mut timer = Timer::syst(cp.SYST, 1.hz(), clocks);
+    let mut timer = Timer::syst(cp.SYST, clocks)
+        .start_count_down(1.hz());
 
     // Wait for the timer to trigger an update and change the state of the LED
     loop {
         block!(timer.wait()).unwrap();
-        led.set_high();
+        led.set_high().unwrap();
         block!(timer.wait()).unwrap();
-        led.set_low();
+        led.set_low().unwrap();
     }
 }
 ```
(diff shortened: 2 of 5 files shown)


TheZoq2:
Ohh, that's a neat feature. The change looks good to me, but the dependency on the void crate should probably be removed from `Cargo.toml`

burrbull:
I could not remove it entirely due to https://github.com/rust-embedded/embedded-hal/blob/a5c0e3c63e4e431807d491d35dd67d1d43c07c92/src/timer.rs#L77

TheZoq2:
Ah right, makes sense. I guess we might want to submit a PR to fix that in embedded-hal at some point.

burrbull:
Yes. But in any case it will be breaking change.

burrbull:
https://github.com/rust-embedded/embedded-hal/pull/150

TheZoq2:
Do we want to merge this now, or wait for the embedded-hal PR to land?

burrbull:
This part of changes shouldn't break user code (I hope).

TheZoq2:
I think it will, we are changing public API after all.

For example, someone might have written wrapper functions around the GPIO pins like this
```rust
fn do_something_with_pins(...) -> Result<Something, Void> {
   pin.set_high()?; // ? Operator no longer works because return type changed from Void to Infallible
}
```

burrbull:
Looks like you are right.
Have you seen this somewhere is used.

TheZoq2:
No, I don't think I have. It also seems like kind of a weird thing to do if you understand the point of the Void type.

It might not make a difference though. We already have breaking changes staged for the next release, so we might as well throw this in. When embedded-hal updates, that too will be a breaking change (probably).

TeXitoi:
Will break https://github.com/TeXitoi/blue-pill-quickstart/blob/master/src/main.rs#L42

TheZoq2:
https://github.com/rust-lang/rust/pull/65355

Looks like the never type will be included in 1.40 :tada:. Do we still want to merge this, and then move to `!` or go straight to `!`? My vote is on the latter, as both things will be breaking changes

burrbull:
By the way is there issue to disable `#[must_use]` check for `Result<smth, !>` to not write `.unwrap()` everywhere?

TheZoq2:
That's a good question, it would certainly be nice

Also, I may have been too quick with my previous message. The comments on that rust-lang issue suggest that there is still some work to be done for it to stabilize.

TheZoq2:
I submitted an issue here: https://github.com/rust-lang/rust/issues/65861

therealprof:
> I don't mind "latest stable" at all, but perhaps someone else has concerns about it. @therealprof What do you think?

I would prefer to stay a couple of versions behind the latest and the greatest. `Infallible` is AFAIK planned to automatically turn into the never type as soon as supported by the compiler so switching to `Infallible` now seems like the right approach to me.

burrbull:
Rebased

TheZoq2:
Great! would you mind moving the changelog entry to the "Brekaing changes" section? Other than that, I think this is ready to be merged

burrbull:
Done.

TheZoq2:
Lovely, thanks!

PR #103, opened 2019-08-18
Remove OutputPin impl for Alternate mode pins

### LINKED ISSUE #102: GPIO output toggling does not work on Alternate pins
I'm using the blue pill board, which has an LED connected to PC13.

With the following code, the LED lights up as soon as the pin is configured as push/pull output, but it cannot be turned off (`set_low` does not turn off the LED, and for sanity checking `set_high` does not either).

```rust
#![deny(unsafe_code)]
#![no_main]
#![cfg_attr(not(test), no_std)]

// Panicking behavior
extern crate panic_semihosting; // logs messages to the host stderr; requires a debugger

#[macro_use]
extern crate cortex_m_rt;

use cortex_m_semihosting::hprintln;
use stm32f1xx_hal::{prelude::*, pac};

#[entry]
fn main() -> ! {
    hprintln!("init").unwrap();

    let device: pac::Peripherals = pac::Peripherals::take().unwrap();

    let mut rcc = device.RCC.constrain();
    let mut gpioc = device.GPIOC.split(&mut rcc.apb2);

    let mut led = gpioc.pc13.into_alternate_push_pull(&mut gpioc.crh);
    led.set_low().unwrap();

    hprintln!("loop").unwrap();
    loop {
    }
}
```

This code uses stm32f1 0.8 and stm32f1xx-hal 0.4. I also tried stm32f1 0.6 and stm32f1xx-hal 0.2, which showed the same behavior.

Any idea what the problem could be?

The full code can be found [here](https://github.com/dbrgn/desklift/tree/led-bug/firmware) (branch `led-bug`).

TheZoq2:
Closes #102

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Breaking changes
 
 - Change timer/pwm init API
+- Remove `set_low` and `set_high` for pins in Alternate output mode
 
 ### Changed
 
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -479,30 +479,6 @@ macro_rules! gpio {
 
                 impl<MODE> toggleable::Default for $PXi<Output<MODE>> {}
 
-                impl<MODE> OutputPin for $PXi<Alternate<MODE>> {
-                    type Error = Void;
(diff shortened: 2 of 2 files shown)


TheZoq2:
Oh yes, good idea. A new commit has been pushed

PR #104, opened 2019-08-26
Add unidirectional SPI DMA

TheZoq2:
This adds basic support for transmitting data over SPI using DMA.

Since this is a basic implementation, only non-circular buffers are supported. However, at the moment I don't have a need for that, and I think we might have to restructure some DMA stuff for simultaneous TX and RX support so I don't want to put too much effort into this until we get around to that.

I think RX support might be fairly easy, but it would require reconfiguring the SPI device as reads happen on writes in bidirectional mode.

*Note*: I have yet to test DMA over SPI2, I'll do that tomorrow but the code probably shouldn't change

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -2,8 +2,9 @@
 
 use core::ptr;
 
-pub use crate::hal::spi::{Mode, Phase, Polarity};
 use nb;
+
+pub use crate::hal::spi::{Mode, Phase, Polarity};
 use crate::pac::{SPI1, SPI2};
 
 use crate::afio::MAPR;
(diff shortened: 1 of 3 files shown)


burrbull:
Can you rebase on master and use AsSlice instead of AsRef?

TheZoq2:
Oops, I guess I forgot to compile after rebasing yesterday 😅

TheZoq2:
> Note: I have yet to test DMA over SPI2, I'll do that tomorrow but the code probably shouldn't change

That took a *bit* longer than a day, but I have finally got around to testing it. Looks like it works which means that this should be ready to merge unless there are any other issues

src/spi.rs
```diff
@@ -226,3 +227,88 @@ hal! {
     SPI1: (_spi1),
     SPI2: (_spi2),
 }
+
+
+
+
```
therealprof: nit: Excess newlines

src/spi.rs
```diff
@@ -226,3 +227,88 @@ hal! {
...
+    TxDma,
+    Transfer,
+    R,
+    Static,
+    TransferPayload,
+};
+
+use core::sync::atomic::{self, Ordering};
+
+use as_slice::AsSlice;
```
therealprof: nit: please move module-wide imports to the top

therealprof:
@TheZoq2 Mind rebasing? Otherwise looks good to go for me.

TheZoq2:
Sure, thanks for the review

PR #106, opened 2019-08-31
Rewrite SWJ config when modifying MAPR (fixes #105)

### LINKED ISSUE #105: MAPR remaping with disable_jtag causes debugger to fail on stm32f103RGT6
I recently got an XL density stm32f103 (stm32f103RGT6) and ran into a pretty strange issue.

When running the following code, the debugger stops responding and I have to reboot the microcontroller with BOOT0 pulled high in order to reflash it. The issue happens when configuring serial, when configuring SPI1 but not when configuring SPI2. 

If `mapr.mapr().modify(|_, w| w.spi1_remap().bit(PINS::REMAP));` is removed from `spi.rs`, things work fine which is a pretty clear indication that it is responsible.

Finally, removing `disable_jtag` also makes things work as normal

```rust
#![deny(unsafe_code)]
#![no_main]
#![no_std]

use panic_semihosting as _;

use stm32f1xx_hal::{
    prelude::*,
    pac,
    serial::{Config, Serial},
    spi::{Spi, Mode, Polarity, Phase}
};
use cortex_m_rt::entry;



#[entry]
fn main() -> ! {
    // Get access to the device specific peripherals from the peripheral access crate
    let p = pac::Peripherals::take().unwrap();

    let mut flash = p.FLASH.constrain();
    let mut rcc = p.RCC.constrain();

    let clocks=rcc.cfgr.freeze(&mut flash.acr);

    // Prepare the alternate function I/O registers
    let mut afio = p.AFIO.constrain(&mut rcc.apb2);

    // Prepare the GPIOB peripheral
    let mut gpiob = p.GPIOB.split(&mut rcc.apb2);


    let (pa15, pb3, pb4) = afio.mapr.disable_jtag(gpioa.pa15, gpiob.pb3, gpiob.pb4);

    /*
    let right_usart = Serial::usart1(
        p.USART1,
        (
            gpioa.pa9.into_alternate_push_pull(&mut gpioa.crh),
            gpioa.pa10,
        ),
        &mut afio.mapr,
        Config::default().baudrate(921_600.bps()),
        clocks,
        &mut rcc.apb2,
    );
    */

    let mode = Mode{
        phase: Phase::CaptureOnFirstTransition,
        polarity: Polarity::IdleLow
    };
    let pins = (
        gpioa.pa5.into_alternate_push_pull(&mut gpioa.crl),
        gpioa.pa6,
        gpioa.pa7.into_alternate_push_pull(&mut gpioa.crl),
    );
    let mode = Mode{
        phase: Phase::CaptureOnFirstTransition,
        polarity: Polarity::IdleLow
    };

    let spi = Spi::spi1(p.SPI1, pins, &mut afio.mapr, mode, 1.khz(), clocks, &mut rcc.apb2);

    /*
    {
        let pins = (
            gpiob.pb13.into_alternate_push_pull(&mut gpiob.crh),
            gpiob.pb14,
            gpiob.pb15.into_alternate_push_pull(&mut gpiob.crh),
        );

        let spi = Spi::spi2(p.SPI2, pins, mode, 1.khz(), clocks, &mut rcc.apb1);
    }
    */
    loop {}
}
```

The same code works perfectly fine on the standard blue pill boards. This might also be a hardware issue, however as far as I know, the debug circuit on my custom PCB is the same as the blue pill.

TheZoq2:
This stack exchange post explains the issue pretty well https://electronics.stackexchange.com/questions/260757/timer1-remap-cause-debug-crash-on-stm32f103

The SWJ_CFG bits are write only, meaning that reads are undefined. On the blue pill, reads seem to work fine anyway, but on the RGT6, they return some other value which messes with the debug port.

My fix for it is to make the `mapr` function private, and only allow modifications via the `modify_mapr` function which remembers the state of the debug port and writes the config accordingly.

Since `mapr` was public, this is a breaking change. I see no reason to have it public, in fact, users can break the abstractions by modifying `mapr` so this new `modify_mapr` function is marked `pub(crate)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Remove `set_low` and `set_high` for pins in Alternate output mode
 - Renames `set_seconds` and `seconds` methods on RTC to `set_time` and `current_time`, respectively
 - Starting the timer does not generate interrupt requests anymore
+- Make MAPR::mapr() private
+
+### Fixed
+
+- Fix some F1 variants crashing when modifying MAPR if JTAG is disabled
 
 ### Changed
 
--- a/src/afio.rs
+++ b/src/afio.rs
@@ -22,7 +22,7 @@ impl AfioExt for AFIO {
 
         Parts {
             evcr: EVCR { _0: () },
-            mapr: MAPR { _0: () },
+            mapr: MAPR { _0: (), jtag_enabled: true },
             exticr1: EXTICR1 { _0: () },
             exticr2: EXTICR2 { _0: () },
             exticr3: EXTICR3 { _0: () },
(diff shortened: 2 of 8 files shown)


geomatsi:
LGTM, thanks !

You probably mean 'SWJ_CFG bits' (not pins) in commit message. FWIW, suggested approach is extensible in terms of support of other debug port mappings. In other words, when more SWJ_CFG combinations are needed, ```jtag_enabled: bool``` can be replaced by appropriate ```enum```.

Regards,
Sergey

TheZoq2:
> FWIW, suggested approach is extensible in terms of support of other debug port mappings

That's true, but since this is internal stuff, I'd say we can make that change if we decide to support other modes

PR #107, opened 2019-09-01
bump as_slice

burrbull:


dvdkon:
`as_slice` 0.2.0 has been yanked from crates.io. Maybe this should be reverted?

PR #108, opened 2019-09-01
public unsafe interface for rcc::Enable/Reset

burrbull:
https://riot.im/app/#/room/#rust-embedded:matrix.org/$hpsOhXJFao1S4JbCH6I_aEy2HFdqXy8f5XCb_d_711E

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Add `InputPin` impl for generic open drain outputs
+
 ## [v0.5.2] - 2019-12-15
 
 - Fix USB module docs
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -5,7 +5,7 @@ use core::marker::PhantomData;
 
 use crate::gpio::Analog;
 use crate::gpio::{gpioa, gpiob, gpioc};
-use crate::rcc::{APB2, Clocks, Enable, Reset};
+use crate::rcc::{APB2, Clocks, sealed::{Enable, Reset}};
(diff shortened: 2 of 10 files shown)


TheZoq2:
Sorry for the delayed reply!

Did I understand the matrix discussion correctly in that this is supposed to be used for using devices that are not supported by the HAL?

If so, I think this looks reasonable, though it would be nice with a comment explaining the reason for this being unsafe, and what happens if you use it incorrectly.

burrbull:
> Did I understand the matrix discussion correctly in that this is supposed to be used for using devices that are not supported by the HAL?

Yes, it is needed for alternative implementations and unsupported peripherals.

It's `unsafe` to indicate that user enable/disable peripheral at his own risk as it could break RCC safety abstraction.

TheZoq2:
Sounds reasonable, though I think that should also be included in the docs for the function

therealprof:
@burrbull Can you adjust the docs so we can get this landed?

burrbull:
I never could write documentation.

burrbull:
And I'm not sure about public interface.
Now it looks like this:
```rust
pub trait Enable {
    unsafe fn enable(apb: &mut Self::Bus);
    unsafe fn disable(apb: &mut Self::Bus);
}
pub trait Reset {
    unsafe fn reset(apb: &mut Self::Bus);
}
```

As these functions are `unsafe` maybe it make sense to simplify it to just:
```rust
pub trait Enable {
    unsafe fn enable();
    unsafe fn disable();
}
pub trait Reset {
    unsafe fn reset();
}
```
and implement for each peripheral?

burrbull:
Or I can split these PR on 2.

Issue #114, opened 2019-10-01
Probably you should rethink API structure


rumatoest:
Hello everyone.

I'm developing software more than 10 years. 
I've seen a lot of anti-patters, but your code really impressed me.
I can definitely say that this is worst API approach I've ever seen in my life.

Looks like you've never try to write anything except POC with your API.
I was trying to create some kind of library using embedded-hal and stm32f1xx-hal 
and found out that it is not worth it.
You should understand that I'm a really stubborn developer and I can dive deep into code 
to understand how it works.

## So what is wrong with your code.
Everything in your type system is wrong.
Sometimes it is even looks like some kind of joke to me, the same as `bool x = true; if (x.toString().length() == 5) {}`

Lets examine :gpio::gpiob::PB0 it is a separate struct that has same api as :gpio::gpiob::PB1
But PB1 is another struct so they can not be swapped in the code easily, you have to rewrite it.
Are you seriously thinking that defining separate types for any possible abstraction is good idea.

Also you've groupped pins into namespaces (probably is is somehow related to ports) and this is where really fun stuff begins.
Somehow deep inside your soul you are understand that using separate struct for each pin is bad idea.
So you've decided to make an abstraction via downgrade method that returns generic pin struct.
So now I'm able to downgrade PB0 to ::gpio::gpioa::PBx which is still bind to its namespace and it is not the same as ::gpio::gpioa::PAx.
Even generic pins in your API are not the same.

Next not so pleasant thing is pin function that mutates it to another "type" like `fn into_push_pull_output(self, cr: &mut CRL) -> PA1<Output<PushPull>>`.
It is not bad to mutate to another abstraction on state change, but only if method signature is like `fn into_push_pull_output(self) -> PA1<Output<PushPull>>`.
Expecting reference to second mutable parameter ruins everything.

My point here is that whole approach to API structure is wrong. 
It is hard to understand and almost impossible to write flexible code.
But why we are using Rust here? We must remove complexity not adding new one.
Right now kernel GPIO API looks much more convenient than yours https://www.kernel.org/doc/Documentation/gpio/consumer.txt

## Should you care about it?
Well I've suggest you too look into Arduino ecosystem. Arduino IDE is terrible and libraries system is half-work solution but it has Wiring with it's so called std library. Some devs clams that wiring is to high level and sometimes to slow comparing to C.
But it is simple, thus may devs were able to start their projects quickly.

[...]

PR #115, opened 2019-10-01
Implement more general GPIO pins

TheZoq2:
This adds the generic pin enum which spun out of #114 

I'm not sure if only having the downgrade function on the already downgraded pins is a good idea as you'd have to do `.downgrade().downgrade()` if you want to go all the way, which I imagine is the thing people want to do most often. Actually, do we even have a use for the half downgraded pins anymore?

This has a fair amount of code duplication, do you think I should make another macro to automatically impl the individual functions. Or should we use the enum_dispatch crate? I'm not a huge fan of adding external dependencies for a single use, mainly for security concerns, but perhaps someone disagrees.

Also, for now this is untested on actual hardware, but I don't see any reason it wouldn't work since it's just an alias.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -217,7 +217,6 @@ macro_rules! gpio {
 
 
 
-
             $(
                 /// Pin
                 pub struct $PXi<MODE> {
@@ -542,6 +541,7 @@ macro_rules! impl_pxx {
                     $(Pxx::$pin(pin) => pin.set_high()),*
                 }
             }
+
             fn set_low(&mut self) -> Result<(), Void> {
                 match self {
                     $(Pxx::$pin(pin) => pin.set_low()),*
@@ -570,6 +570,7 @@ macro_rules! impl_pxx {
                     $(Pxx::$pin(pin) => pin.is_high()),*
                 }
             }
+
             fn is_low(&self) -> Result<bool, Void> {
                 match self {
                     $(Pxx::$pin(pin) => pin.is_low()),*
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -95,6 +95,7 @@ macro_rules! gpio {
                 State,
                 Active,
                 Debugger,
+                Pxx
             };
 
             /// GPIO parts
@@ -154,6 +155,12 @@ macro_rules! gpio {
             pub struct $PXx<MODE> {
                 i: u8,
                 _mode: PhantomData<MODE>,
+            }
+
+            impl<MODE> $PXx<MODE> {
+                pub fn downgrade(self) -> Pxx<MODE> {
+                    Pxx::$PXx(self)
+                }
             }
 
             impl<MODE> OutputPin for $PXx<Output<MODE>> {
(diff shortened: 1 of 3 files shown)


therealprof:
At some point I was hoping we could get rid of this horrid macro mess. I'd probably rather use `enum_dispatch` which I just saw for the first time and it's kind of love on first sight. 😄
I don't mind either way though, if people like this -- let's land it.

TeXitoi:
Do you have any idea to remove thé macro?

I'll try to review this PR today.

therealprof:
@TeXitoi 

> Do you have any idea to remove thé macro?

No idea, still toying around with some type system magic to find a solution for that. However the more complex the macro the more work it is to get rid of it later.

TheZoq2:
> As it would be a breaking change to just return Pxx on downgrade, maybe we can add some deprecated type definition type PAx<MODE> = Pxx<MODE> to be backward compatible. It will remove the access of the real PAx pin, but I'm not really sure that's useful.

This sounds like a good solution to me

Oh, and I noticed that the pins implement `InputPin` for `Output<OpenDrain>` so that should probably be added here before the merge

TeXitoi:
@therealprof I'll try to look at the macro an see if it can be rewritten using something similar to https://github.com/jamwaffles/embedded-graphics/blob/master/embedded-graphics/src/fonts/font_builder.rs

TheZoq2:
I had another look at `enum_dispatch` and it looks like it requires annotating the relevant traits which might not be possible in this case since those come from `embedded_hal`.

TheZoq2:
Could be, I'll investigate later today

-------- Original Message --------

TheZoq2:
Should be fixed now. I also added an example to make sure it actually works, as I previously forgot to re-add the `downgrade` function.

Any further comments? @therealprof @TeXitoi

src/gpio.rs
```diff
@@ -206,6 +213,11 @@ macro_rules! gpio {
                 }
             }
 
+            pub type $PXx<MODE> = Pxx<MODE>;
+
+
+
```
therealprof:
```suggestion
```

src/gpio.rs
```diff
@@ -507,6 +524,69 @@ macro_rules! gpio {
...
+        }
+
+        impl<MODE> OutputPin for Pxx<Output<MODE>> {
+            type Error = Void;
+            fn set_high(&mut self) -> Result<(), Void> {
+                match self {
+                    $(Pxx::$pin(pin) => pin.set_high()),*
+                }
+            }
+            fn set_low(&mut self) -> Result<(), Void> {
```
therealprof:
```suggestion

            fn set_low(&mut self) -> Result<(), Void> {
```

src/gpio.rs
```diff
@@ -507,6 +524,69 @@ macro_rules! gpio {
...
+        }
+
+        impl<MODE> InputPin for Pxx<Input<MODE>> {
+            type Error = Void;
+            fn is_high(&self) -> Result<bool, Void> {
+                match self {
+                    $(Pxx::$pin(pin) => pin.is_high()),*
+                }
+            }
+            fn is_low(&self) -> Result<bool, Void> {
```
therealprof:
```suggestion

            fn is_low(&self) -> Result<bool, Void> {
```

therealprof:
I only have a few whitespace nits, other than that (and possibly the distribution over 5 commits) that looks really good.

Issue #42, opened 2019-03-10
PWM expose 2 channels, take their pins and force their mode.


nraynaud:
There are a few issues with the current code: 
 - timers only expose 2 channels
 - timers takes the channels pins wether we want it or not.
 - timers force the pin mode to be OpenDrain or PushPull

example form timer_input:
```rust
impl Pins<TIM1> for (PA8<Alternate<OpenDrain>>, PA9<Alternate<OpenDrain>>) {
    const REMAP: u8 = 0b00;
}

```

I propose having the channels pins using the `Option` and a generic mode:
```rust
pub type Pwm1Mapping0<M1, M2, M3, M4> = (Option<PA8<Alternate<M1>>>,
                                         Option<PA9<Alternate<M2>>>,
                                         Option<PA10<Alternate<M3>>>,
                                         Option<PA11<Alternate<M4>>>);
```

here is a more complete example:
https://github.com/japaric/stm32f103xx-hal/blob/9f132cd3a8a95d0b726f925c239fad6e74b60d76/src/pwm2.rs#L219

TheZoq2:
This looks like it has been fixed by allowing external impls of the `Pins` trait. This was documented in #60 

However, I feel like it might be more ergonomic to impl all the possible pin/channel combinations, probably via a macro. Is there a reason we haven't done that before?

geomatsi:
Hi @TheZoq2 ,
At first I tried to come up with macro, but did not succeeded. So I switched to newtype approach. IIUC some clever macro still would be an option here. My only concern is firmware size: is it possible to list all the combination without bloating the final binary ?

Regards,
Sergey

TeXitoi:
Only thé used code Will be on the binary so it shouldn't be a problem.

TheZoq2:
I'll try to experiment with this when I get time

TheZoq2:
This has been fixed in #147

Issue #118, opened 2019-10-07
Support for different SPI configurations


katyo:
This is awesome HAL in terms of usage safety than any other I seen before. But it isn't so yet flexible in several cases.

I would like to configure SPI for display driver, which uses unidirectional setup with only SCLK and MOSI lines.

As I can see, the Pins trait implementation for that case is missing.
Since MISO pin in target device already used as GPIO for other purpose, so I cannot use Pins<(sclk, miso, mosi)>.

Also there are some other SPI setups like a bi-directional half-duplex with single data line.

therealprof:
Hey @katyo, thanks for this great suggestion. Indeed the implementation is rather barebones and lacking a lot of functionality. Unidirectional communication should be rather easy to implement by defining dummy pins and allowing their use like it has been done e.g. here: https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/spi.rs .

Maybe you'd like to give it a try?

katyo:
Okay, it looks not so difficult.

TheZoq2:
This is something I would find useful as well!

Let us know if you run into any issues if you decide to implement it!

Windfisch:
Hi!
I made [an attempt at implementing this](https://github.com/Windfisch/stm32f1xx-hal/commit/06f097afa96ac1798d872fe4bd3528167ea95c9e), but it turned out more difficult than expected.
The issue is that we cannot just go [the stm32f4xx-hal way](https://github.com/stm32-rs/stm32f4xx-hal/blob/2ed3d3569543ba5edb2b8cc3338aa23af2d0dd73/src/spi.rs#L371,L388) by implementing traits `PinMiso<SPI>`, `PinMosi<SPI>` and `PinSck<SPI>`, all of which contain the `NoMiso`/etc. types in addition to the actual pins, and then requiring a tuple of these three traits.
This works for the STM32F4, but not with the remapping system of the F1...

The only other way I found is to have eight `impl pins<SPI1> for (..., ..., ...) { blah = true }`, where the `...` enumerate all possible combinations of pin and not-a-pin, and then have eight more for the blah=false case, and then eight more for SPI2 :/

Do you have a better idea on how to do this?

Windfisch:
[...]

PR #120, opened 2019-10-11
RFC: Macro for generating all PWM map combinations

TheZoq2:
This is a new solution to #42 

It is currently not working because it produces conflicting impls for some cases, in particular it will produce several impls for `((), (), (), ())` and for mappings where parts are shared, such as `((), (), PB0, PB1)`

I could try and come up with a solution for those two problems, but the macro is "a bit" ugly, so I would first like to see if this is something we are interested in before spending too much time improving it.

Right now I'm using `()` as a "don't care" marker, but that should probably be changed. Either to `pwm::Unused`, or just using a tuple with fewer elements.


src/pwm.rs
```diff
@@ -153,54 +154,130 @@ pub trait Pins<TIM> {
...
+impl_pin_combinations!(PA0,  PA1, PA2,  PA3; for TIM2,  remap 0b00);
+impl_pin_combinations!(PA15, PB3, PA2,  PA3; for TIM2,  remap 0b01);
+impl_pin_combinations!(PA0,  PA1, PB10, PB11; for TIM2, remap 0b01);
+impl_pin_combinations!(PA15, PB3, PB10, PB11; for TIM2, remap 0b01);
+
+impl_pin_combinations!(PA6, PA7, PB0, PB1; for TIM3, remap 0b00);
+impl_pin_combinations!(PB4, PB5, PB0, PB1; for TIM3, remap 0b01);
+impl_pin_combinations!(PC6, PC7, PC8, PC9; for TIM3, remap 0b11);
+
+impl_pin_combinations!(PB6,  PB7,  PB8,  PB9 ; for TIM4, remap 0b0);
```
geomatsi: Is there any reason you want to keep 'for' here ?

src/pwm.rs
```diff
@@ -153,54 +154,130 @@ pub trait Pins<TIM> {
...
+impl_pin_combinations!(PD12, PD13, PD14, PD15; for TIM4, remap 0b1);
```
geomatsi: Could you please also update pwm examples. It would be good to see how those pre-defined combinations are used e.g. in pwm_custom.rs example.

src/pwm.rs
TheZoq2: Not really, I just feelt like it makes it slightly more readable

src/pwm.rs
TheZoq2: Yea, that should probably be done. Though it's essentially just passing a tuple as pins, i.e.

```rust
let (ch0, ch2, ch3) = Pwm::new(..., (pb6, (), pb8, pn9)
```

geomatsi: Am I correct assuming that providing incorrect tuple of pins ends up in compilation error ?

TheZoq2: Yep :)

TheZoq2:
After seeing the proposed solution to #118, I decided that just enumerating the possible options in a macro is probably easier than the recursive abomination I had before. Especially since some pins are shared between different remaps.

The PR has been updated to reflect that. The usage should remain the same, and before we merge, I should rewrite the examples and docs.

TheZoq2:
I'll close this in favour of #124

Issue #35, opened 2019-02-25
EXTI interrupts


gbip:
Is there any way to setup EXTI interrupts at the moment ?
I need to have an interrupt handler when a rising front occurs on a GPIO pin, how could I do that ?

TheZoq2:
There is, but I don't remember the exact way to do it. In one of the crates, probably cortex-m, there is a macro to define an interrupt, it takes a function along with this enum https://docs.rs/stm32f1/0.6.0/stm32f1/stm32f103/enum.Interrupt.html that specifies which interrupt to listen for.

Using interrupts without cortex-m-rtfm is a bit annoying because passing ownership of peripherals into interrupt handlers is impossible, but it is useable with some unsafe code

gbip:
Thank you !
I don't see anything to set up the gpio side of things. Is it not yet implemented ? I might take a look.

therealprof:
I've used EXTI on STM32 before here: https://github.com/stm32-rs/stm32f0xx-hal/blob/master/examples/led_hal_button_irq.rs

Just no one has made it into something HAL-worthy.

NB: It is possible to pass ownership of peripherals safely but yeah, it's annoying.

gbip:
I will try to come up with a proposition, otherwise I will just use your code as inspiration, thanks !

TeXitoi:
I reopen this issue because I think that's an important feature request.

gbip:
I was thinking about the [following](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=de39eaf7dbe704cbd67f630cb2f6677e) API to attach handlers to certain events.

*Pros* :
* easy to use
* use the type system to enforce safe concurrency

*Cons* :
* I have no idea how to setup the interrupt handler in a transparent way for the crate user

TeXitoi:
@gbip you can't do that at zero cost. You'll have to stock the closure somewhere, and you don't know it's size ahead of time.

Better to first provide an interface close to the hardware but safe as a first version.

therealprof:
We do have a (non-exhaustive) implementation to setup EXTI on pins in `stm32f4xx-hal`: https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/gpio.rs

tib888:
I believe this will be the ultimate solution for interrupts and safety is already under development by @japaric: https://github.com/japaric/cortex-m-rtfm

However it would be nice to see an example made for STM32F103 EXTI...

[...]

PR #125, opened 2019-10-23
ExtiPin trait for GPIOs

Windfisch:
This addresses #35 by porting the ExtiPin trait from the https://github.com/stm32-rs/stm32f4xx-hal repo. (It already contains my recent fix that was discussed over there yesterday.)

Before merging, I would be very happy about a short review :).
If you'd like me to squash the commits together, please tell me.


burrbull:
LGTM. But it will be even better if you add example in `example` dir.

src/gpio.rs
```diff
@@ -188,6 +210,69 @@ macro_rules! gpio {
...
+                            afio.exticr3.exticr3().modify(|r, w| unsafe {
+                                w.bits((r.bits() & !(0xf << offset)) | ($extigpionr << offset))
+                            });
+                        },
+                        12..=15 => {
+                            afio.exticr4.exticr4().modify(|r, w| unsafe {
+                                w.bits((r.bits() & !(0xf << offset)) | ($extigpionr << offset))
+                            });
+                        },
+                        _ => {}
```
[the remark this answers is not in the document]
Windfisch: Good point, thanks. Since this was copypasted from stm32f4xx-hal, it should probably changed there and in stm32h7x-hal.

Windfisch:
> LGTM. But it will be even better if you add example in `example` dir.

Good point, I'll add one.

src/gpio.rs
```diff
@@ -518,6 +538,51 @@ macro_rules! gpio {
...
+                            exticr = (exticr & !(0xf << offset)) | ($extigpionr << offset);
+                            w.bits(exticr)
+                        });
+                    }
+
+                    /// Generate interrupt on rising edge, falling edge or both
+                    fn trigger_on_edge(&mut self, exti: &mut EXTI, edge: Edge) {
+                        match edge {
+                            Edge::RISING => {
+                                exti.rtsr.modify(|r, w| unsafe { w.bits(r.bits() | (1 << $i)) });
```
geomatsi: There already exist access methods for all the registers. For instance, example for EXTI1 w/o macros woud look like this: 
```
        exti.imr.modify(|_, w| w.mr1().set_bit());
        exti.ftsr.modify(|_, w| w.tr1().set_bit());
        exti.rtsr.modify(|_, w| w.tr1().set_bit());
```
Is it possible to use the same access methods in order to avoid the use of ```unsafe``` block ?

src/gpio.rs
```diff
@@ -269,7 +269,7 @@ macro_rules! gpio {
 
                 /// Clear the interrupt pending bit for this pin
                 fn clear_interrupt_pending_bit(&mut self, exti: &EXTI) {
-                    exti.pr.modify(|r, w| unsafe { w.bits(r.bits() | (1 << self.i)) });
+                    exti.pr.modify(|_r, w| unsafe { w.bits(1 << self.i) });
```
geomatsi: Maybe just ```_``` instead of ```_r```: 
```
exti.pr.modify(|_, w| unsafe { w.bits(1 << self.i) });
```

src/gpio.rs
```diff
@@ -518,6 +538,51 @@ macro_rules! gpio {
...
                 }
+
+                impl<MODE> ExtiPin for $PXi<Input<MODE>> {
+                    /// Configure EXTI Line $i to trigger from this pin.
+                    fn make_interrupt_source(&mut self, afio: &mut AFIO) {
+                        let offset = 4 * ($i % 4);
+                        afio.$exticri.modify(|r, w| unsafe {
+                            let mut exticr = r.bits();
+                            exticr = (exticr & !(0xf << offset)) | ($extigpionr << offset);
+                            w.bits(exticr)
```
geomatsi: There already exist access methods for ```AFIO_EXTICR``` registers. For instance, example for EXTI10_15 w/o macros woud look like this:
```
        // select PD15 as source input for line EXTI10_15
        afio.exticr4
            .exticr4()
            .modify(|_, w| unsafe { w.exti15().bits(0b0011) });
```
So is it possible to simplify suggested implementation a bit using ```exti``` accessors ?

src/gpio.rs
Windfisch: That's a good point, I was thinking about that myself.
However, I don't think using the mr1/mr2/mr... registers is feasible, since that would make it very difficult to generate such code in the first place. Currently, the macro for the pins just gets a pin number and with that, it is able to set the proper bits for imr, ftsr and rtsr (and other stuff outside of my patch).
Doing anything else would make the macro invocations much more lengthy and, more importantly, error-prone. It would be easy to slip in a copy-paste-mistake where pin3 uses mr3, but tr2 by accident.

Also, since this is inside of a library, I don't see it as a problem here to use `unsafe`. All in all I'd say it's not worth the hassle.

src/gpio.rs
Windfisch: that part should have been replaced by `.write` already.

src/gpio.rs
Windfisch: The same as for the `imr, rtsr, ftsr` registers applies here: It would be possible to use these accessors, but at the cost of having to repeat the pin number a couple of times in each of the n*16 pin definitions (n being the number of ports), as `imrN`, `rtsrN`, `exticrM` and so on.
I don't think it would simplify things.

src/gpio.rs
TheZoq2: Sounds reasonable to me

geomatsi: Sounds reasonable to me as well. Thanks for clarification!

geomatsi:
I would also suggest to add an example. Otherwise LGTM. Thanks !

Windfisch:
I've added an example. Additionally I implemented a `check_interrupt()` function that checks whether this EXTI pin has actually caused an interrupt. While this is obvious for EXTI1..4, this check is required for EXTI5-9 (which all share one ISR handler).

I'm open for suggestions about the `check_interrupt()` function, the name doesn't yet feel right.

examples/exti.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+}
+
+#[entry]
+fn main() -> ! {
+    // initialization phase
+    let p = pac::Peripherals::take().unwrap();
+    let cp = cortex_m::peripheral::Peripherals::take().unwrap();
+    {
+        // the scope ensures that the exti reference is dropped before the first ISR can be executed.
```
TheZoq2: I'm assuming you do this to avoid a race condition when between `main` and the interrupt handler.

However, I don't think *holding* a reference to something can trigger a race condition, that can only happen if you actually use the reference.

Or is this more of a warning to users that they shouldn't use exti after nvic.enable?

Windfisch: It' both. It's a warning, and also: Isn't it UB that two mutable references to the same data *exist*? (Regardless of them being used?)

TheZoq2: I'm not entirely sure. My guess is that it's not for the same reason that working with pointers is safe until you actually dereference them, which requires `unsafe` but I may be wrong

TeXitoi: With non lexical lifetime, it's safe if you don't use the mutable reference before enabling interrupt, as it would be accepted by the compiler to create another `&mut` after the enable line.

For examples, I prefer a static mut option to a static mut maybeuninit, less unsafe things to think about.

examples/exti.rs
```diff
@@ -0,0 +1,68 @@
...
+
+        let mut rcc = p.RCC.constrain();
+        let mut gpioa = p.GPIOA.split(&mut rcc.apb2);
+        let mut gpioc = p.GPIOC.split(&mut rcc.apb2);
+        let mut afio = p.AFIO.constrain(&mut rcc.apb2);
+
+        let led = unsafe { &mut *LED.as_mut_ptr()};
+        *led = gpioc.pc13.into_push_pull_output(&mut gpioc.crh);
```
TeXitoi:
```rust
let led = gpioc.pc13.into_push_pull_output(&mut gpioc.crh);
unsafe { LED = MaybeUninit::new(led) };
```

TeXitoi: no need of scope with this.

examples/exti.rs
```diff
@@ -0,0 +1,68 @@
...
+
+        let exti = unsafe { &mut *EXTI.as_mut_ptr()};
+        *exti = gpioa.pa7.into_floating_input(&mut gpioa.crl);
+        exti.make_interrupt_source(&mut afio);
+        exti.trigger_on_edge(&p.EXTI, Edge::RISING_FALLING);
+        exti.enable_interrupt(&p.EXTI);
```
TeXitoi: Same pattern:
```rust
let exti = gpioa.pa7.into_floating_input(&mut gpioa.crl);
exti.make_interrupt_source(&mut afio);
exti.trigger_on_edge(&p.EXTI, Edge::RISING_FALLING);
exti.enable_interrupt(&p.EXTI);
unsafe { EXTI = MaybeUninit::new(exti) };
```

TheZoq2:
@Windfisch This is pretty close to getting merged, would you mind fixing the new conflicts.

Also, a changelog entry would be nice :)

Windfisch:
thanks for the feedback, @TheZoq2. I'll have a look at it

Windfisch:
@TheZoq2, I've resolved the merge conflict and added a Changelog entry.

Do you prefer me to squash together all my changes, or merge it as is? (Or can you just check a "squash" checkbox in Github?)

TheZoq2:
Thanks!

There is a nice big "Squash and merge" button that I can use :)

TheZoq2:
@Windfisch Is there a reason you used a SCREAMING_SNAKE_CASE name for the `Rising` enum? The compiler now gives a warning for that, so if there isn't a particular reason for it, I would like to change it

Windfisch:
Hi @TheZoq2, I took most of the code from the stm32f4xx-hal implementation, and they use SCREAMING_SNAKE_CASE, too: 
https://github.com/stm32-rs/stm32f4xx-hal/blob/e94c88ec85488445bbef10542e21173a99781364/src/gpio.rs#L77
For compatibility reasons, I suggest to keep it the same in all hals.

TheZoq2:
Ah alright, i'll add a #[allow(...)] to that enum then :)

Issue #127, opened 2019-10-26
Some i2c devices do not work correctly


JPFrancoia:
Hi, I'd like to use a HTU21D humidity sensor with a blue pill. Unfortunately, this sensor doesn't have an "official" driver on this page: https://github.com/rust-embedded/awesome-embedded-rust/blob/master/README.md#driver-crates.

But I found *one* repository implementing the driver: https://github.com/etrombly/htu21d.
But I can't get it to work. Here is the code I'm using:

```rust
// #![deny(unsafe_code)]
// #![deny(warnings)]
#![no_main]
#![no_std]

extern crate cortex_m as cm;
extern crate cortex_m_rt as rt;
extern crate embedded_hal;
extern crate nb;
extern crate panic_itm;
extern crate stm32f1xx_hal as hal;
extern crate htu21d;
// extern crate itoa;

use nb::block;
use rt::entry;

// use embedded_hal::digital::v2::OutputPin;
use hal::prelude::*;
use hal::{
    i2c, pac,
    serial::{Config, Serial},
    delay::Delay,
};

use htu21d::Htu21df;

#[entry]
fn main() -> ! {
    // Initialize peripherals
    let device_peripherals = pac::Peripherals::take().unwrap();
    let cortex_peripherals = cortex_m::Peripherals::take().unwrap();

    let mut rcc = device_peripherals.RCC.constrain();
    let mut flash = device_peripherals.FLASH.constrain();
    let clocks = rcc.cfgr.freeze(&mut flash.acr);

    // Initialize the GPIO A and B
    // let mut gpioa = device_peripherals.GPIOA.split(&mut rcc.apb2);
    let mut gpiob = device_peripherals.GPIOB.split(&mut rcc.apb2);

    // Prepare the alternate function I/O registers
    let mut afio = device_peripherals.AFIO.constrain(&mut rcc.apb2);

    let scl = gpiob.pb6.into_alternate_open_drain(&mut gpiob.crl);
    let sda = gpiob.pb7.into_alternate_open_drain(&mut gpiob.crl);

    let i2c = i2c::BlockingI2c::i2c1(
        device_peripherals.I2C1,
        (scl, sda),
        &mut afio.mapr,
        i2c::Mode::Standard { frequency: 400 },
        clocks,
        &mut rcc.apb1,
        50000,
        3,
        10000,
        50000,
    );

    let mut htu21d = Htu21df::new(i2c).unwrap();
[...]

PR #128, opened 2019-10-27
Set i2c frequency using Hertz

TheZoq2:
In #127 I realised that i2c doesn't use the Hertz structs to set freuency. This seems strange, so I fixed it.

Unlike the other peripherals, this doesn't allow construction from `KiloHertz` or `MegaHertz` as that would have required making the `mode` struct generic, which would propagate a lot. It might still be a good idea to do so, though perhaps it's better to pass the frequency separately to the i2c "constructor". This would require less generics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Renames `set_seconds` and `seconds` methods on RTC to `set_time` and `current_time`, respectively
 - Starting the timer does not generate interrupt requests anymore
 - Make MAPR::mapr() private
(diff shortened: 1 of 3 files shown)


TeXitoi:
Any blocker on this or can it be merged?

TheZoq2:
I think it can be merged, unless we want to allow `KiloHertz` and `MegaHertz`

TeXitoi:
True, `Into<Hertz>` as in `rcc.CFGR` could be great.

TheZoq2:
I added methods to construct Mode from `impl Into<Hertz>` to avoid spreading generics everywhere and still give the option to construct from other frequency types. @TeXitoi Any conceerns with that implementation?

PR #130, opened 2019-11-02
Changed timer stop_in_debug to modify dbg.cr instead of overwriting it

cs2dsb:
Quick fix to prevent configuring pwm_input clobbering the dbg.cr register.

The issue I was having was ITM was breaking after the call to stop_in_debug in timer.rs because part of the ITM config lives in dbg.cr and the call to write was resetting everything to 0.

I had a quick scan for other similar issues in timer.rs but didn't see any.


TheZoq2:
Good catch, thanks for the PR!

PR #132, opened 2019-11-08
restore motor example

burrbull:


Cargo.toml
```diff
@@ -69,9 +69,8 @@ version = "1.5.2"
 #git = "https://github.com/japaric/jnet"
 #rev = "df96b408049ca952ad7844d6552e87cf8fc18d2a"
 
-#[dev-dependencies.motor-driver]
-#git = "https://github.com/japaric/motor-driver"
-#rev = "00c8b15223643090d69f1acfb8b7a7a43a440417"
+[dev-dependencies.motor-driver]
+git = "https://github.com/japaric/motor-driver"
```
therealprof: Is this release-safe? I'm not not sure about the current state of git dev dependencies, but regular git dependencies prevent publishing to crates.io.

TheZoq2:
I don't really see the point in this example, it seems to mostly showcase the use of peripherals that are used in other examples as well.

therealprof:
Well, if we don't fix it we should get rid of it. All the dead code around doesn't look too exciting.

PR #99, opened 2019-08-17
adc-dma timer

burrbull:
cc @TheZoq2 
cc @therealprof

PR #133, opened 2019-11-09
Added all timers for all variants as described by CubeMX

cs2dsb:
I've added all the timers according to the CubeMX database with the correct device & density flags.

The only one's I've had to comment out are 9, 10, 12, 13 and 14 on some devices because fields for these are missing from the underlying pac. The code is there to enable them when stm32-rs fixes up those SVDs.

This might be a minor breaking change as previously some of the devices had timers enabled that aren't actually available across the whole range - the fix is to add the appropriate density feature. I've added some info about this to the README so hopefully people will find this without a problem.

There's also some minor formatting stuff mostly from rustfmt in case you are wondering about the whitespace changes in pwm_input.

I should also draw your attention to the fact I added #![allow(unused_unsafe)] (with a comment) to lib.rs to hide the extremely numerous warnings that were coming up because some pac have safe fields incorrectly identified as unsafe. I think adding #![allow(unused_unsafe)] to the crate is fine and preffererable to it being harder to read the compiler output but I'm happy to be persuaded otherwise if you disagree.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,14 +94,21 @@
 
 #![no_std]
 
+// Some pac crates have fields specified in such a way that they are safe and other
+// have them unsafe (likely an SVD error that needs patching). Unsafe blocks for 
+// one device cause warnings for the safe devices. This disables that warning.
+#![allow(unused_unsafe)]
+
 // If no target specified, print error message.
 #[cfg(not(any(feature = "stm32f100", feature = "stm32f101", feature = "stm32f103")))]
 compile_error!("Target not found. A `--feature <target-name>` is required.");
 
 // If any two or more targets are specified, print error message.
-#[cfg(all(feature = "stm32f100", feature = "stm32f101"))]
-#[cfg(all(feature = "stm32f100", feature = "stm32f103"))]
-#[cfg(all(feature = "stm32f101", feature = "stm32f103"))]
+#[cfg(any(
+    all(feature = "stm32f100", feature = "stm32f101"),
+    all(feature = "stm32f100", feature = "stm32f103"),
+    all(feature = "stm32f101", feature = "stm32f103"),
+))]
 compile_error!("Multiple targets specified. Only a single `--feature <target-name>` can be specified.");
 
 #[cfg(feature = "device-selected")]
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -269,7 +269,7 @@ impl CFGR {
             _ => (true, false),
         };
 
-        let apre_bits = self
+        let apre_bits: u8 = self
             .adcclk
             .map(|adcclk| match pclk2 / adcclk {
                 0..=2 => 0b00,
@@ -530,17 +530,14 @@ macro_rules! ahb_bus {
     }
 }
 
-#[cfg(any(
-    feature = "stm32f100",
-    feature = "stm32f103",
-))]
-bus! {
-    TIM1 => (APB2, tim1en, tim1rst),
-}
 #[cfg(feature = "stm32f103")]
 bus! {
     ADC2 => (APB2, adc2en, adc2rst),
     CAN1 => (APB1, canen, canrst),
+}
+#[cfg(all(feature = "stm32f103", feature = "high",))]
+bus! {
+    ADC3 => (APB2, adc3en, adc3rst),
 }
 bus! {
     ADC1 => (APB2, adc1en, adc1rst),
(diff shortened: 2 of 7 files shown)


src/timer.rs
```diff
@@ -269,16 +292,75 @@ macro_rules! hal {
...
     TIM4: (tim4, pclk1_tim, dbg_tim4_stop),
 }
+
+#[cfg(feature = "HIGH")]
+hal! {
+    TIM5: (tim5, pclk2_tim, dbg_tim5_stop),
+}
+
+/*
+    dbg_tim(12-13)_stop fields missing from 103 xl in stm32-rs
```
geomatsi: Maybe leave FIXME or TODO mark here and in the next commented section to remind about the required work ?

src/rcc.rs
```diff
@@ -80,13 +78,10 @@ impl APB1 {
...
 }
 
-
 impl APB1 {
     /// Set power interface clock (PWREN) bit in RCC_APB1ENR
     pub fn set_pwren(&mut self) {
-        self.enr().modify(|_r, w| {
-            w.pwren().set_bit()
-        })
+        self.enr().modify(|_r, w| w.pwren().set_bit())
```
geomatsi: It looks like your commit includes ```cargo fmt``` formatting changes. IMHO it makes sense to keep coding style changes in a separate commit just to simplify review.

src/timer.rs
cs2dsb: Will do.

src/rcc.rs
cs2dsb: I'll back these out and update the PR shortly.

src/timer.rs
cs2dsb: Done and pushed

src/rcc.rs
cs2dsb: Done and pushed

cs2dsb:
I left the piddling little rustfmt changes from random files out but added the pwm_input formatting fix back in as a separate commit because it took me half an hour to tweak the indentation and I feel like it's worth keeping ;)

src/timer.rs
```diff
@@ -269,16 +290,74 @@ macro_rules! hal {
...
+}
+
+#[cfg(feature = "medium")]
 hal! {
-    TIM2: (tim2, pclk1_tim, dbg_tim2_stop),
-    TIM3: (tim3, pclk1_tim, dbg_tim3_stop),
     TIM4: (tim4, pclk1_tim, dbg_tim4_stop),
 }
+
+#[cfg(feature = "HIGH")]
```
burrbull: Why uppercase here?

cs2dsb: Good spot, that's a mistake.

cs2dsb: Fixed and pushed.

src/timer.rs
```diff
@@ -1,7 +1,28 @@
 //! # Timer
 
 use crate::hal::timer::{CountDown, Periodic};
-use crate::pac::{DBGMCU as DBG, TIM1, TIM2, TIM3, TIM4};
+use crate::pac::{DBGMCU as DBG, TIM2, TIM3};
+#[cfg(any(feature = "stm32f100", feature = "stm32f103", feature = "stm32f105",))]
+use crate::pac::TIM1;
+#[cfg(feature = "medium")]
+use crate::pac::TIM4;
+#[cfg(feature = "HIGH")]
```
burrbull: Same here.

cs2dsb: Fixed and pushed

cs2dsb:
Rebased latest changes from master.

Is there anything else I need to address to get this merged? I'm going to have a go at implementing the various PWM input & output features for the advanced timers and want to make sure that PR is easy to merge too.

Thanks
Daniel

src/timer.rs
```diff
@@ -269,16 +307,91 @@ macro_rules! hal {
...
+
+#[cfg(all(
+    feature = "stm32f103", 
+    feature = "high",
+))]
+hal! {
+    TIM8: (tim8, pclk2_tim, dbg_tim8_stop),
+}
+
+//TODO: restore these timers once stm32-rs has been updated
```
burrbull: `stm32-rs` 0.9 released and we now use it or you mean something other?

cs2dsb: That refers to the missing dbg_tim(9 to 14)_stop fields on dbgmcu. They haven't been implemented in stm32-rs yet.

src/lib.rs
```diff
@@ -94,14 +94,21 @@
 
 #![no_std]
 
+// Some pac crates have fields specified in such a way that they are safe and other
+// have them unsafe (likely an SVD error that needs patching). Unsafe blocks for 
+// one device cause warnings for the safe devices. This disables that warning.
+#![allow(unused_unsafe)]
```
TheZoq2: I think this error should have been fixed by the latest PAC release, could you check that?

cs2dsb: There are still instances of mixed safe & unsafe registers. Examples:
stm32f100::tim1::aar::AAR_W::bits is safe
stm32f100::tim6::aar::AAR_W::bits is unsafe

I think the original comment still stands for now.

burrbull:
I think it can be merged after conflicts to be resolved.

cs2dsb:
Updated with latest master.
Note I had to modify the new start_master (#99) part of the macro because the general purpose timers 9 to 14 don't support master mode.

src/lib.rs
TheZoq2: Depending on how many such warnings there are, I would prefer having the `#[allow]` thing on the lines/functions where it happens. That way we don't miss actually useful warnings

PR #137, opened 2019-11-17
Add rudimentary docs for the usb module

TheZoq2:
@Disasm Since you wrote the module, could you make sure I didn't make any obvious misstakes here?

### DIFF
--- a/src/usb.rs
+++ b/src/usb.rs
@@ -1,6 +1,6 @@
 //! USB peripheral
 
-//! Requires the `stm32_usbd` feature.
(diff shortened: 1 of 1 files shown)


src/usb.rs
```diff
@@ -1,5 +1,9 @@
 //! USB peripheral
 
+//! Requires the `stm32_usbd` feature.
```
Disasm:
```suggestion
//! Requires the `stm32-usbd` feature.
```

PR #140, opened 2019-11-21
Removed crate level allow(unused_unsafe) in favour of attributes on the specific instances

### LINKED ISSUE #139: More granular #[allow(unused_unsafe)]
In #133 we added `#![allow(unused_unsafe)]` due to differences between devices in the PAC crate. Since those warnings are useful in general, we should probably move the `#[allow]` declarations to the places where there are actual problems rather than making it global

cs2dsb:
Resolves #139 

I temporary added `#![deny(unused_unsafe)] and built all variants to find the instances of unused unsafe. Turned out there were only two left after the recent PAC update.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,11 +94,6 @@
 
 #![no_std]
 
-// Some pac crates have fields specified in such a way that they are safe and other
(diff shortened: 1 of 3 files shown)


TheZoq2:
Lovely, thanks!

PR #141, opened 2019-11-22
Flash Read/Write Support

luctius:
This pr adds support for flash reading and writing.

I think the API can be improved and I am open for suggestions on that front.

### DIFF
--- a/src/flash.rs
+++ b/src/flash.rs
@@ -2,6 +2,242 @@
 
 use crate::pac::{flash, FLASH};
 
+pub const FLASH_START: u32 = 0x0800_0000;
+pub const FLASH_END:   u32 = 0x080F_FFFF;
+
+const RDPRT_KEY: u16 = 0x00A5;
+const KEY1: u32 = 0x45670123;
+const KEY2: u32 = 0xCDEF89AB;
+
+pub const SZ_1K: u16 = 1024;
+
+pub type Result<T> = core::result::Result<T, Error>;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
+pub enum Error {
+    Success,
+    AddressLargedThanFlash,
+    AddressMisAligned,
+    LengthNotMultiple2,
+    LengthTooLong,
+    EraseError,
+    ProgrammingError,
+    WriteError,
+    VerifyError,
+    UnLockError,
+    LockError,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
+pub enum SectorSize {
+    Sz1K = 1,
+    Sz2K = 2,
+    Sz4K = 4,
+}
+impl SectorSize {
+    fn size(self) -> u16 {
+        SZ_1K * self as u16
+    }
+}
+
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
+pub enum FlashSize {
+    Sz16K  = 16,
+    Sz32K  = 32,
+    Sz64K  = 64,
+    Sz128K = 128,
+    Sz256K = 256,
+    Sz384K = 384,
+    Sz512K = 512,
+    Sz768K = 768,
+    Sz1M   = 1024,
+}
+impl FlashSize {
+    fn size(self) -> u32 {
+        SZ_1K as u32 * self as u32
+    }
+}
+
+pub struct FlashWriter<'a> {
+    flash: &'a mut Parts,
+    sector_sz: SectorSize,
+    flash_sz: FlashSize,
+}
+impl<'a> FlashWriter<'a> {
+    fn unlock(&mut self) -> Result<()> {
+        // Wait for any ongoing operations
+        while self.flash.sr.sr().read().bsy().bit_is_set() {}
+
+        // NOTE(unsafe) write Keys to the key register
+        unsafe { self.flash.keyr.keyr().write(|w| w.key().bits(KEY1)); }
+        unsafe { self.flash.keyr.keyr().write(|w| w.key().bits(KEY2)); }
+
+        while self.flash.sr.sr().read().bsy().bit_is_set() {}
+
+        // Verify success
+        match self.flash.cr.cr().read().lock().bit_is_clear() {
+            true => Ok(()),
+            false => Err(Error::UnLockError),
+        }
+    }
+
+    fn lock(&mut self) -> Result<()> {
+        while self.flash.sr.sr().read().bsy().bit_is_set() {}
+
+        self.flash.cr.cr().modify(|_, w| w.lock().set_bit());
+
+        while self.flash.sr.sr().read().bsy().bit_is_set() {}
+
+        match self.flash.cr.cr().read().lock().bit_is_set() {
+            true => Ok(()),
+            false => Err(Error::LockError),
+        }
+    }
+
+    fn valid_address(&self, offset: u32) -> Result<()> {
+        if FLASH_START + offset > FLASH_END {
+            Err(Error::AddressLargedThanFlash)
+        }
+        else if offset % 2 != 0 {
+            Err(Error::AddressMisAligned)
+        }
+        else {
+            Ok(())
+        }
+    }
+
+    fn valid_length(&self, length: u32) -> Result<()> {
+        if length > self.sector_sz.size() as u32 {
+            Err(Error::LengthTooLong)
+        }
+        else if length % 2 != 0 {
+            Err(Error::LengthNotMultiple2)
+        }
+        else {
+            Ok(())
+        }
+    }
+
+    pub fn page_erase(&mut self, start_offset: u32) -> Result<()> {
+        self.valid_address(start_offset)?;
+
+        self.unlock()?;
+
+        // Set Page Erase
+        self.flash.cr.cr().modify(|_, w| w.per().set_bit() );
+
+        // Write address bits
+        // NOTE(unsafe) This sets the sector address in the Address Register
+        unsafe { self.flash.ar.ar().write(|w| w.far().bits(FLASH_START + start_offset) ); }
+
+        // Start Operation
+        self.flash.cr.cr().modify(|_, w| w.strt().set_bit() );
+
+        // Wait
+        while self.flash.sr.sr().read().bsy().bit_is_set() { }
+
+        // Check for errors
+        let sr = self.flash.sr.sr().read();
+
+        // Remove Page Erase Operation bit
+        self.flash.cr.cr().modify(|_, w| w.per().clear_bit() );
+
+        // Re-lock flash
+        self.lock()?;
+
+        for idx in 0..self.sector_sz.size() -1 {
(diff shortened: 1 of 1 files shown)


src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+            true => Ok(()),
+            false => Err(Error::LockError),
+        }
+    }
+
+    fn valid_address(&self, offset: u32) -> Result<()> {
+        if FLASH_START + offset > FLASH_END {
+            Err(Error::AddressLargedThanFlash)
+        }
+        else if offset % 2 != 0 {
```
thalesfragoso: I wonder if the compiler will always do the right thing and change this to `offset & 1`, otherwise it might be better if we use bitwise AND instead of mod.

src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+        else {
+            Ok(())
+        }
+    }
+
+    fn valid_length(&self, length: u32) -> Result<()> {
+        if length > self.sector_sz.size() as u32 {
+            Err(Error::LengthTooLong)
+        }
+        else if length % 2 != 0 {
```
thalesfragoso: Same as above

src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+        // Check for errors
+        let sr = self.flash.sr.sr().read();
+
+        // Remove Page Erase Operation bit
+        self.flash.cr.cr().modify(|_, w| w.per().clear_bit() );
+
+        // Re-lock flash
+        self.lock()?;
+
+        for idx in 0..self.sector_sz.size() -1 {
```
thalesfragoso: It might be good to add a way to ignore verification, like a `verify: bool` argument.

src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+            let write_address = (FLASH_START + start_offset + idx as u32) as *mut u16;
```
thalesfragoso: Don't you have to correct this to not go pass a section boundary ? i.e `start_offset` might be in the middle of a sector. Also I don't think it needs to be `mut`.

src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+            let verify: u16 = unsafe { core::ptr::read_volatile(write_address) };
+            if verify != 0xFFFF {
+                return Err(Error::VerifyError);
+            }
+        }
+        if sr.wrprterr().bit_is_set() {
```
thalesfragoso: Why not test this before verification ? Also, I think you need to reset the bit before returning.

src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+    /// Retrieve an slice of at most 1 sector of data from FLASH_START +offset
+    pub fn read(&mut self, offset: u32, length: usize) -> Result<&'static [u8]> {
+        self.valid_address(offset)?;
+        self.valid_length(length as u32)?;
+
+        let address = (FLASH_START + offset) as *const _;
+
+        // NOTE(unsafe) grants read access to the FLASH
+        unsafe {
+            Ok(core::slice::from_raw_parts::<'static, u8>(address, length))
```
thalesfragoso: Is this safe ? Doing that you give a static immutable reference, but you can modify its contents with `write`. Am I missing something ?

src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+        }
+
+        for offset in (start_offset..start_offset + length as u32).step_by(self.sector_sz.size() as usize) {
+            self.page_erase(offset)?;
+        }
+
+        Ok(())
+    }
+
+    /// Retrieve an slice of at most 1 sector of data from FLASH_START +offset
```
thalesfragoso: Why have this limitation ?

src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+
+        let address = (FLASH_START + offset) as *const _;
+
+        // NOTE(unsafe) grants read access to the FLASH
+        unsafe {
+            Ok(core::slice::from_raw_parts::<'static, u8>(address, length))
+        }
+    }
+
+    /// Write at most 1 sector of data to FLASH_START +offset
```
thalesfragoso: Why have this limitation ?

src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+
+            // Wait for write
+            while self.flash.sr.sr().read().bsy().bit_is_set() {}
+
+            // Set Page Programming to 0
+            self.flash.cr.cr().modify(|_, w| w.pg().clear_bit() );
+
+            // Check for errors
+            let sr = self.flash.sr.sr().read();
+            if self.flash.sr.sr().read().pgerr().bit_is_set() {
```
thalesfragoso: Reset bit if set ?

src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+                self.lock()?;
+                return Err(Error::ProgrammingError);
+            }
+            else if self.flash.sr.sr().read().wrprterr().bit_is_set() {
```
thalesfragoso: Same as above

src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+                self.lock()?;
+                return Err(Error::WriteError);
+            }
+
+            // Verify written word
+            let verify: u16 = unsafe { core::ptr::read_volatile(write_address) };
```
thalesfragoso: Add option to avoid verification ?

src/flash.rs
luctius: I didn't think it would be writeable, but if it is, then it is indeed a problem. Got a suggestion on an alternative?

luctius:
You make excellent points. It was a while since I really worked on the code and although I gave it a final look-over I apparently missed quite a bit.

Thanks for the thorough check.

Most of it is easily fixed which I will probably do in the coming week.

src/flash.rs
thalesfragoso: You can't write through the reference, but you can use `FlashWriter::write` to modify the same memory that you're holding a reference to.
Hmm, maybe you can return a proxy type to prevent someone to call write and then use the ref. Maybe something like:
```rust
struct FlashReader<'a> {
    _flash: &'a FlashWriter,
    data: &'a [u8],
}

impl<'a> FlashReader<'a> {
    pub fn data(&self) -> &[u8] {
        self.data
    }
}
```
Then, you return that on the `FlashWriter::read` method.

And other things that I missed, I think `read` should only ask for `&self` instead of `&mut self` and also, have you tested if we can read a byte of a time instead of two ? The `Read Operation` part of the manual doesn't mention anything, so I think we can, but I didn't test it or read the entire manual.

src/flash.rs
```diff
@@ -2,6 +2,242 @@
...
+    Sz64K  = 64,
+    Sz128K = 128,
+    Sz256K = 256,
+    Sz384K = 384,
+    Sz512K = 512,
+    Sz768K = 768,
+    Sz1M   = 1024,
+}
+impl FlashSize {
+    fn size(self) -> u32 {
```
TheZoq2: I would prefer a more descriptive name for this, perhaps `as kbytes`

src/flash.rs
TheZoq2: I had a look at the LLVM-IR produced by rustc, and it looks like it doesn't change it to an `and` in debug mode, but does in release mode. That is without LLVM optimizations though

src/flash.rs
```diff
@@ -34,3 +307,94 @@ impl ACR {
         unsafe { &(*FLASH::ptr()).acr }
     }
 }
+
+/// Opaque AR register
+pub struct AR {
+    _0: (),
+}
+
+#[allow(dead_code)]
```
TheZoq2: Why are these allow(dead_code) needed? Is rustc missing the usages, or are they currently unused?

luctius: Those are currently unused; I added them for completeness.

luctius:
The comments made should have been fixed, though perhaps some unsafe code comments could still be better.

I have verified that single byte read operations from flash are possible.

src/flash.rs
```diff
@@ -2,6 +2,275 @@
...
+    sector_sz: SectorSize,
+    flash_sz: FlashSize,
+    verify: bool,
+}
+impl<'a> FlashWriter<'a> {
+    fn unlock(&mut self) -> Result<()> {
+        // Wait for any ongoing operations
+        while self.flash.sr.sr().read().bsy().bit_is_set() {}
+
+        // NOTE(unsafe) write Keys to the key register
```
TheZoq2: Still doesn't explain why unsafe is safe

src/flash.rs
```diff
@@ -2,6 +2,275 @@
...
+        self.valid_address(start_offset)?;
+
+        // Unlock Flash
+        self.unlock()?;
+
+        // Set Page Erase
+        self.flash.cr.cr().modify(|_, w| w.per().set_bit() );
+
+        // Write address bits
+        // NOTE(unsafe) This sets the sector address in the Address Register
```
TheZoq2: Same here

src/flash.rs
```diff
@@ -2,6 +2,275 @@
...
+
+        // Unlock Flash
+        self.unlock()?;
+
+        for idx in (0..data.len() ).step_by(2) {
+            self.valid_address(offset + idx as u32)?;
+
+            let hword: u16 =
+                  (data[idx]   as u16)
+                | (data[idx+1] as u16) << 8;
```
TheZoq2: Weird formatting. Though I'm not sure what the best way to format it is

irwineffect:
@luctius, @TheZoq2, if the remaining comments are addressed, would this be ready to merge in?

TheZoq2:
It has been a while since I looked at it, but IIRC, yes

src/flash.rs
burrbull: it's better to use `copy_from_slice` and `u16::from_be/le_bytes` for such manipulations

DerFetzer:
@luctius Any news on this? Is there help needed somewhere?

irwineffect:
I had been working on addressing the comments (https://github.com/irwineffect/stm32f1xx-hal/commit/0a997b0aa870b26f9f59eeb0d037aa49d4439f4a), although when I started to do more testing of the branch I had noticed some issues when erasing the flash, I think the erase logic as it exists right now isn't quite correct (the reference manual PM0075 also is somewhat ambiguous). It's been a while since I had looked at it, but I'll dig up what I had been working on and post more details.

irwineffect:
Well, I've been testing and I haven't been able to replicate the issues I had seen earlier, everything seems ok...perhaps my hardware was just wigging out? Sorry for the doubt, @luctius.

irwineffect:
It seems that @luctius hasn't signed on in a long time...should I create a pull request for irwineffect@0a997b0? This is based on @luctius's branch, plus I've tried to address the comments.

TheZoq2:
Sorry for the late reply. 

> should I create a pull request for irwineffect/stm32f1xx-hal@0a997b0? This is based on @luctius's branch, plus I've tried to address the comments.
I'd say you should, it's pointless to have a stale PR waiting when one that can make progress is available :)

irwineffect:
This PR can now be continued on #257

PR #145, opened 2019-11-25
alternative implementation for SPI pins

burrbull:


therealprof:
Looks good.

burrbull:
What is better?
```rust
let spi = dp.SPI2.spi(
        (spi::NoSck, spi::NoMiso, gpioa.pa10),
...
```
or 
```rust
let spi = dp.SPI2.spi(
        (gpio::NoPin, gpio::NoPin, gpioa.pa10),
...
```

therealprof:
```
let spi = dp.SPI2.spi(
        (spi::NoSck, spi::NoMiso, gpioa.pa10),
```

Is better. It expresses intent and thus prevents misconfiguration.

burrbull:
Ok.

But misconfiguration is not possible in second variant too.

therealprof:
> But misconfiguration is not possible in second variant too.

Well, misconfiguration in the sense of the user accidentally opting out of the wrong thing. If you read `NoPin` you know that you're not using a pin but which one. If you only allow `NoMiso` then everyone can see on a first glance that you're not using a physical MISO pin.

TheZoq2:
I agree that `NoMiso/NoMosi` is better than `NoPin`.

I would almost like to change it to `SpiPins{miso, mosi, sck}` for even more clarity. I find myself always having to look up the ordering of the pins when using SPI. That is more verbose though, so we should probably allow both.

Perhaps we could add `Impl<Into>

TheZoq2:
Looks good in its current state, we can probably do the thing I mentioned in the previous comment later.

burrbull:
In last commit I've implemented `Pins` trait for different order.
Now you can pass `(sck, miso, mosi)` or `(mosi, miso, sck)` or `(mosi, NoMiso, NoSck)`, etc.

TheZoq2:
Looks good, thanks!

PR #146, opened 2019-11-26
Pwm remap

burrbull:
Same as #145

### DIFF
--- a/examples/pwm_custom.rs
+++ b/examples/pwm_custom.rs
@@ -8,37 +8,13 @@ use panic_halt as _;
 
 use cortex_m::asm;
 use stm32f1xx_hal::{
-    gpio::gpiob::{PB4, PB5},
-    gpio::{Alternate, PushPull},
-    pac::{self,TIM3},
+    pac,
     prelude::*,
-    pwm::{Pins, Pwm, C1, C2},
     timer::Timer,
+    gpio::NoPin,
 };
 
 use cortex_m_rt::entry;
-
-// Using PB5 channel for TIM3 PWM output 
-// struct MyChannels(PB5<Alternate<PushPull>>);
-// impl Pins<TIM3>  for MyChannels {
-//     const REMAP: u8 = 0b10;
-//     const C1: bool = false;
-//     const C2: bool = true;
-//     const C3: bool = false;
-//     const C4: bool = false;
-//     type Channels = Pwm<TIM3, C2>;
-// }
-
-// Using PB4 and PB5 channels for TIM3 PWM output
-struct MyChannels(PB4<Alternate<PushPull>>, PB5<Alternate<PushPull>>);
-impl Pins<TIM3> for MyChannels {
-    const REMAP: u8 = 0b10;
-    const C1: bool = true;
-    const C2: bool = true;
-    const C3: bool = false;
-    const C4: bool = false;
-    type Channels = (Pwm<TIM3, C1>, Pwm<TIM3, C2>);
-}
 
 #[entry]
 fn main() -> ! {
@@ -59,11 +35,7 @@ fn main() -> ! {
     let p1 = gpiob.pb5.into_alternate_push_pull(&mut gpiob.crl);
 
     let mut pwm = Timer::tim3(p.TIM3, &clocks, &mut rcc.apb1)
-        .pwm(
-            MyChannels(p0, p1),
-            &mut afio.mapr,
-            1.khz(),
-        );
+        .pwm((p0, p1, NoPin, NoPin), &mut afio.mapr, 1.khz());
 
     let max = pwm.0.get_max_duty();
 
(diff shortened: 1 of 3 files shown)


geomatsi:
How does this play with #124 ?

examples/pwm_custom.rs
```diff
@@ -59,11 +35,7 @@ fn main() -> ! {
     let p1 = gpiob.pb5.into_alternate_push_pull(&mut gpiob.crl);
 
     let mut pwm = Timer::tim3(p.TIM3, &clocks, &mut rcc.apb1)
-        .pwm(
-            MyChannels(p0, p1),
-            &mut afio.mapr,
-            1.khz(),
-        );
+        .pwm((p0, p1, NoPin, NoPin), &mut afio.mapr, 1.khz());
```
geomatsi: IIUC you have somehow to specify Tim3PartialRemap here.

src/pwm.rs
```diff
@@ -138,117 +138,159 @@ use crate::pac::{TIM2, TIM3, TIM4};
...
+        }
+    }
 }
 
+remap!(Tim2NoRemap, TIM2, 0b00, PA0, PA1, PA2, PA3);
+remap!(Tim2PartialRemap1, TIM2, 0b01, PA15, PB3, PA2, PA3);
+remap!(Tim2PartialRemap2, TIM2, 0b10, PA0, PA1, PB10, PB11);
+remap!(Tim2FullRemap, TIM2, 0b11, PA15, PB3, PB10, PB11);
+remap!(Tim3NoRemap, TIM3, 0b00, PB4, PB5, PB0, PB1);
+remap!(Tim3PartialRemap, TIM3, 0b10, PA6, PA7, PB0, PB1);
```
geomatsi: TIM3 remap values look incorrect: PB4, PB5 correspond to 0b10

examples/pwm_custom.rs
burrbull: You are right. In some cases it's needed to specify remap:
```rust
    let mut pwm = Timer::tim3(p.TIM3, &clocks, &mut rcc.apb1)
        .pwm::<Tim3PartialRemap, _, _>((NoPin, NoPin, pb0, NoPin), &mut afio.mapr, 1.khz());
```

But not in this one.

geomatsi: Could you please clarify how it is going to work in this case ? How proper remap will be figured out from the specified list of 2 pins ?

burrbull: For TIM3 CH1 and CH2 there is only one combination of pins for each remap and compiler can infer type of remap.
But if you specify only PB0 and/or PB1 (CH3 and CH4) compiler don't know what remap do you mean (`Tim3NoRemap` or `Tim3PartialRemap`).

geomatsi: Ok, thanks. 

BTW, doc section in src/pwm.rs needs to be modified as well since now there is no need to use newtype pattern.

src/pwm.rs
```diff
@@ -138,117 +138,159 @@ use crate::pac::{TIM2, TIM3, TIM4};
...
-    const C4: bool = true;
-    type Channels = (Pwm<TIM2, C1>, Pwm<TIM2, C2>, Pwm<TIM2, C3>, Pwm<TIM2, C4>);
+impl<TIM, REMAP, CH1, CH2, CH3, CH4> Pins<REMAP> for (CH1, CH2, CH3, CH4)
+where
+    REMAP: Remap<Periph = TIM>,
+    CH1: Ch1<REMAP>,
+    CH2: Ch2<REMAP>,
+    CH3: Ch3<REMAP>,
+    CH4: Ch4<REMAP> {
+    type Channels = (Pwm<TIM, C1>, Pwm<TIM, C2>, Pwm<TIM, C3>, Pwm<TIM, C4>);
```
geomatsi: Unless I am missing something, C1-C4 channel specifiers become useless after your changes, since channel enablement is done based on chX_state()

burrbull:
In fact this implementation has one problem, which I don't know how to solve on stable Rust (without macros like in #124).
`.pwm()` always returns 4-element tuple.

src/pwm.rs
burrbull: `NoPin` has `false` value.

examples/pwm_custom.rs
geomatsi: There is one thing that worth mentioning in document section. Previous implementation returned tuple of activated channels, the type ```Channels``` was defined as a part of newtype pattern in each case. Now tuple of 4 channels will be returned in all the cases. So you have to remember which channels are not activated based on where ```NoPin``` are specified.

src/pwm.rs
geomatsi: Correct. But then we don't need C1-C4 phantoms.

burrbull:
Closed in favour of #147

PR #148, opened 2019-12-02
Documentation updates and cleanup

### LINKED ISSUE #144: pwm example compile error
```
error[E0599]: no function or associated item named `tim4` found for type `stm32f1xx_hal::timer::Timer<_>` in the current scope
  --> examples/pwm.rs:49:26
   |
49 |     let mut pwm = Timer::tim4(p.TIM4, &clocks, &mut rcc.apb1)
   |                          ^^^^
   |                          |
   |                          function or associated item not found in `stm32f1xx_hal::timer::Timer<_>`
   |                          help: there is a method with a similar name: `tim1`

error: aborting due to previous error
```
on f103

#133 ?

TheZoq2:
This fixes #144, both in terms of making that example compile without extra features, and documenting the features.

While i'm at it, I also updated the PWM and timer docs, and cleaned up `Cargo.toml`

### DIFF
--- a/src/pwm.rs
+++ b/src/pwm.rs
@@ -21,9 +21,9 @@
       gpioa.pa3.into_alternate_push_pull(&mut gpioa.crl),
   );
 
-  // Set up the timer as a PWM output. Since there are multiple remap
-  // options for tim2 that use the same pins, you need to specify
-  // the remap generic parameter.
+  // Set up the timer as a PWM output. If selected pins may correspond to different remap options,
+  // then you must specify the remap generic parameter. Otherwise, if there is no such ambiguity,
+  // the remap generic parameter can be omitted without complains from the compiler.
   let (c1, c2, c3, c4) = Timer::tim2(p.TIM2, &clocks, &mut rcc.apb1)
       .pwm::<Tim2NoRemap, _, _, _>(pins, &mut afio.mapr, 1.khz())
       .3;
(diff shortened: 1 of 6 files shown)


src/pwm.rs
```diff
@@ -13,12 +13,24 @@
...
   let pins = (
       gpioa.pa0.into_alternate_push_pull(&mut gpioa.crl),
       gpioa.pa1.into_alternate_push_pull(&mut gpioa.crl),
       gpioa.pa2.into_alternate_push_pull(&mut gpioa.crl),
       gpioa.pa3.into_alternate_push_pull(&mut gpioa.crl),
   );
+
+  // Set up the timer as a PWM output. Since there are multiple remap
+  // options for tim2 that use the same pins, you need to specify
+  // the remap generic parameter.
```
geomatsi: cc: @burrbull 

In fact, this is not exactly so. It is possible to omit remap generic parameter in many cases when compiler is able to deduce mapping from the selected pins. It can be skipped on the first attempt and compiler will complain if it can not figure out the specific pin combination :)

How about updating this section, e.g. something along the following lines:

```
Set up the timer as a PWM output. If selected pins may correspond to different remap options,
then you shall specify the remap generic parameter. Otherwise, if there is no such ambiguity,
the remap generic parameter can be omitted without complains from the compiler.
```

TheZoq2: Yea, that's what I meant to say but I agree it's pretty unclear. Your wording is a lot better :)

PR #150, opened 2019-12-06
Fixed I2C reading problems

Okarin99:
This PR fixes some issues I had with I2C.
I wanted to query exactly 2 bytes over I2C, but I got an I2C arbitration error.
Because of this I fixed the code to follow exactly the rules described [here](http://st.com/content/ccc/resource/technical/document/application_note/5d/ae/a3/6f/08/69/4e/9b/CD00209826.pdf/files/CD00209826.pdf/jcr:content/translations/en.CD00209826.pdf).
Now it is working fine for me.

Example code:
```rust
#![no_std]
#![no_main]

use cortex_m::iprintln;
use cortex_m_rt::entry;
use mpu6050::*;
use stm32f1xx_hal::{delay::Delay, i2c, prelude::*, stm32};

#[allow(unused_imports)]
use panic_itm;

#[entry]
fn main() -> ! {
    let device = stm32::Peripherals::take().unwrap();
    let mut cp = cortex_m::Peripherals::take().unwrap();
    let stim = &mut cp.ITM.stim[0];

    let mut flash = device.FLASH.constrain();
    let mut rcc = device.RCC.constrain();

    let clocks = rcc.cfgr.freeze(&mut flash.acr);
    let mut afio = device.AFIO.constrain(&mut rcc.apb2);

    let mode = i2c::Mode::Standard {
        frequency: 100_000.hz(),
    };

    let mut gpiob = device.GPIOB.split(&mut rcc.apb2);

    let scl_pin = gpiob.pb8.into_alternate_open_drain(&mut gpiob.crh);
    let sda_pin = gpiob.pb9.into_alternate_open_drain(&mut gpiob.crh);

    let delay = Delay::new(cp.SYST, clocks);

    let i2c_val = i2c::BlockingI2c::i2c1(
        device.I2C1,
        (scl_pin, sda_pin),
        &mut afio.mapr,
        mode,
        clocks,
        &mut rcc.apb1,
        50000,
        10,
        50000,
        50000,
    );

    let mut mpu = Mpu6050::new(i2c_val, delay);
    mpu.init().unwrap();
    mpu.soft_calib(Steps(100)).unwrap();
    mpu.calc_variance(Steps(50)).unwrap();

    iprintln!(stim, "Calibrated with bias: {:?}", mpu.get_bias().unwrap());
    iprintln!(
        stim,
        "Calculated variance: {:?}",
        mpu.get_variance().unwrap()
    );

    loop {
        /*
        // Didn't work before
[...]

PR #152, opened 2019-12-13
fix frequency for SPI2/3

burrbull:


src/i2c.rs
```diff
@@ -253,12 +253,12 @@ macro_rules! hal {
                     pins: PINS,
                     mode: Mode,
                     clocks: Clocks,
-                    apb: &mut APB1,
```
TheZoq2: Is this a breaking change?

If I understand it correctly, it shouldn't be even though the type of a public function is changing

src/rcc.rs
```diff
@@ -634,3 +664,8 @@ bus! {
     TIM10 => (APB2, tim10en, tim10rst),
     TIM11 => (APB2, tim11en, tim11rst),
 }
+
+#[cfg(any(feature = "stm32f102", feature = "stm32f103"))]
+bus! {
+    USB => (APB1, usben, usbrst),
```
TheZoq2: This should probably be used by the USB code, right?

src/i2c.rs
burrbull: No it is not breaking change (it is still `APB1`).

burrbull:
> Looks good for the most part, I added two comments in the code.
> 
> Also, would you mind adding a changelog entry. Or perhaps that's not required since most of this doesn't touch public API apart from fixing the wrong clock being used for SPI

I think it make sense to release 0.5.1 with this patch, because SPI2 and SPI3 work with incorrect frequency for now.

TheZoq2:
> I think it make sense to release 0.5.1 with this patch, because SPI2 and SPI3 work with incorrect frequency for now.

Agreed, i'll do that tomorrow after also reviewing #150

TheZoq2:
v0.5.1 (and v0.5.2 because I failed at git) have now been released with this change included :)

PR #161, opened 2020-01-03
Use RegisterBlock for SPI & I2C instead of macro & make rcc::Bus sealed

burrbull:
~~Depends on #108~~

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -59,8 +59,6 @@
   |    RX    |     PB11      |     PC11           |      PD9        |
 */
 
-
-use core::ops::Deref;
 use core::marker::PhantomData;
 use core::ptr;
 use core::sync::atomic::{self, Ordering};
@@ -79,11 +77,7 @@ use crate::gpio::gpiod::{PD5, PD6, PD8, PD9};
 use crate::gpio::{Alternate, Floating, Input, PushPull};
 use crate::rcc::{RccBus, Clocks, GetBusFreq, sealed::{Enable, Reset}};
 use crate::time::{U32Ext, Bps};
-
-pub trait RegisterBlock {
-    type RB;
-    unsafe fn rb() -> *const Self::RB;
-}
+use crate::utils::RegisterBlock;
 
 /// Interrupt event
 pub enum Event {
@@ -296,14 +290,6 @@ macro_rules! hal {
                     Serial::<$USARTX, _, _>::_usart(usart, pins, config, clocks, apb)
                 }
             }
-
-            impl RegisterBlock for $USARTX {
-                type RB = crate::serial::UsartRegisterBlock;
-                #[inline(always)]
-                unsafe fn rb() -> *const Self::RB {
-                    $USARTX::ptr()
-                }
-            }
         )+
     }
 }
@@ -337,7 +323,7 @@ type UsartRegisterBlock = crate::pac::usart1::RegisterBlock;
 
 impl<USART, REMAP, PINS> Serial<USART, REMAP, PINS>
 where
-    USART: Deref<Target = UsartRegisterBlock> + RccBus + Enable + Reset,
+    USART: RegisterBlock<RB = UsartRegisterBlock> + RccBus + Enable + Reset,
     USART::Bus: GetBusFreq,
 {
     fn _usart(
@@ -352,12 +338,12 @@ where
         USART::reset(apb);
 
         // enable DMA transfers
-        usart.cr3.write(|w| w.dmat().set_bit().dmar().set_bit());
+        unsafe { &*USART::rb() }.cr3.write(|w| w.dmat().set_bit().dmar().set_bit());
 
         // Configure baud rate
         let brr = USART::Bus::get_frequency(&clocks).0 / config.baudrate.0;
         assert!(brr >= 16, "impossible baud rate");
-        usart.brr.write(|w| unsafe { w.bits(brr) });
+        unsafe { &*USART::rb() }.brr.write(|w| unsafe { w.bits(brr) });
 
         // Configure parity and word length
         // Unlike most uart devices, the "word length" of this usart device refers to
@@ -369,7 +355,7 @@ where
             Parity::ParityEven => (true, true, false),
             Parity::ParityOdd => (true, true, true),
         };
-        usart.cr1.modify(|_r, w| {
+        unsafe { &*USART::rb() }.cr1.modify(|_r, w| {
             w
                 .m().bit(word_length)
                 .ps().bit(parity)
@@ -383,14 +369,14 @@ where
             StopBits::STOP2 => 0b10,
             StopBits::STOP1P5 => 0b11,
         };
-        usart.cr2.modify(|_r, w| {
+        unsafe { &*USART::rb() }.cr2.modify(|_r, w| {
             w.stop().bits(stop_bits)
         });
 
         // UE: enable USART
         // RE: enable receiver
         // TE: enable transceiver
-        usart
+        unsafe { &*USART::rb() }
             .cr1
             .modify(|_r, w| w.ue().set_bit().re().set_bit().te().set_bit());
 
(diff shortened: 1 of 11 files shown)


therealprof:
Lovely!

burrbull:
I have removed dependency from 108. Let's merge this first.

burrbull:
Fix commit name.

geomatsi:
Hi @burrbull ,

Could you please provide some brief high level explanation of this PR ? In particular, which problems are you solving using
* ```mod sealed``` for ```RccBus```
* ```Deref``` for both ```I2C``` and ```SPI```.

Regards,
Sergey

burrbull:
I made `Enable` and `Reset` sealed, because I plan to do public unsafe wrappers for these traits.
`RccBus` still public.

`Deref` are needed to make generic implementations without macros and make this `hal` more consistent with others.

No changes in public API for `SPI` and `I2C`.

TeXitoi:
I find this refactor great. I'm just not sure Deref is an appropriate trait for this, why not using a dedicated trait?

burrbull:
> I find this refactor great. I'm just not sure Deref is an appropriate trait for this, why not using a dedicated trait?

What you mean by `dedicated`?

TeXitoi:
```rust
pub(crate) trait RegisterBlock {
    type Type;
    fn register_block(&self) -> *const Target;
}
```

TeXitoi:
From https://doc.rust-lang.org/std/ops/trait.Deref.html

> Deref should only be implemented for smart pointers to avoid confusion.

src/serial.rs
```diff
@@ -52,9 +77,14 @@ use crate::gpio::gpiob::{PB10, PB11, PB6, PB7};
 use crate::gpio::gpioc::{PC10, PC11};
 use crate::gpio::gpiod::{PD5, PD6, PD8, PD9};
 use crate::gpio::{Alternate, Floating, Input, PushPull};
-use crate::rcc::{RccBus, Clocks, Enable, Reset, GetBusFreq};
+use crate::rcc::{RccBus, Clocks, GetBusFreq, sealed::{Enable, Reset}};
 use crate::time::{U32Ext, Bps};
 
+pub trait RegisterBlock {
+    type RB;
+    unsafe fn rb() -> *const Self::RB;
```
burrbull: like this? :)

TeXitoi: Yep, and without `Deref` bounds in the code.

TeXitoi: Note that the method is not unsafe, only dereferencing the returned pointer is unsafe, so no need of unsafe in this trait.

burrbull: Ok. If this trait will common, where (what module) it is better to locate it?

TeXitoi: it should go in stm32hal-utils crate ;-)

Waiting for such a project, I'd say a `pub(crate) mod utils;` at the top level.

burrbull: Can I ask question? Why you don't like `Deref`?

TeXitoi: See https://github.com/stm32-rs/stm32f1xx-hal/pull/161#issuecomment-570601881

TeXitoi: Moreover, the deref implementation will be available to the user, and, for me, it looks more like an implementation detail than a functionnality.

geomatsi:
> I made `Enable` and `Reset` sealed, because I plan to do public unsafe wrappers for these traits.
> `RccBus` still public.
> 
> `Deref` are needed to make generic implementations without macros and make this `hal` more consistent with others.
> 
> No changes in public API for `SPI` and `I2C`.

Thanks for educating me. Makes total sense now.

Regards,
Sergey

src/serial.rs
burrbull: Done.

burrbull:
In my view it is reasonable to generate RegisterBlock trait implementation by `svd2rust` same as `Deref`.

src/serial.rs
```diff
@@ -78,40 +102,65 @@ pub enum Error {
...
-    const REMAP: u8;
+mod sealed {
+    pub trait Remap {
+        type Periph;
+        const REMAP: u8;
+    }
+    pub trait TxPin<REMAP> {}
+    pub trait RxPin<REMAP> {}
+    pub struct _Tx;
+    pub struct _Rx;
```
TheZoq2: This could use some documentation. Suggestion:

```rust
    /// Implemented for a struct which defines a possible remaping
    /// of the `Periph` peripheral
    pub trait Remap {
        type Periph;
        const REMAP: u8;
    }
    /// Implemented for the gpio pins that can be used as TX with a specific
    /// Remaping
    pub trait TxPin<REMAP> {}
    pub trait RxPin<REMAP> {}
    /// NOTE: Not sure what these do, care to explain?
    pub struct _Tx;
    pub struct _Rx;
```

src/serial.rs
```diff
@@ -78,40 +102,65 @@ pub enum Error {
...
 }
+use sealed::{Remap, TxPin, RxPin};
 
-impl Pins<USART1> for (PA9<Alternate<PushPull>>, PA10<Input<Floating>>) {
-    const REMAP: u8 = 0;
+pub trait Pins<REMAP, P> {
+    type _Pos;
```
TheZoq2: What does `_Pos` mean here?

src/serial.rs
```diff
@@ -78,40 +102,65 @@ pub enum Error {
...
+        }
+        impl TxPin<$name> for $TX<Alternate<PushPull>> {}
+        impl RxPin<$name> for $RX<Input<Floating>> {}
+    }
 }
 
-impl Pins<USART3> for (PD8<Alternate<PushPull>>, PD9<Input<Floating>>) {
-    const REMAP: u8 = 0b11;
-}
+// TODO: check density
```
TheZoq2: Unfixed TODO

src/serial.rs
```diff
@@ -221,7 +271,7 @@ macro_rules! hal {
                 /// `MAPR` and `APBX` are register handles which are passed for
                 /// configuration. (`MAPR` is used to map the USART to the
                 /// corresponding pins. `APBX` is used to reset the USART.)
-                pub fn $usartX(
+                pub fn $usartX<POS>(
```
TheZoq2: POS is also uncelar here.

src/serial.rs
```diff
@@ -78,40 +102,65 @@ pub enum Error {
...
+    pub trait TxPin<REMAP> {}
+    pub trait RxPin<REMAP> {}
+    pub struct _Tx;
+    pub struct _Rx;
 }
+use sealed::{Remap, TxPin, RxPin};
 
-impl Pins<USART1> for (PA9<Alternate<PushPull>>, PA10<Input<Floating>>) {
-    const REMAP: u8 = 0;
+pub trait Pins<REMAP, P> {
```
TheZoq2: Could also use a comment, something like

```rust
/// Combination of TX and RX pins useable with the specified Remap
```

src/serial.rs
```diff
@@ -230,204 +280,271 @@ macro_rules! hal {
...
+                    PINS: Pins<REMAP, POS>,
                 {
-                    // enable and reset $USARTX
-                    $USARTX::enable(apb);
-                    $USARTX::reset(apb);
-
                     #[allow(unused_unsafe)]
                     mapr.modify_mapr(|_, w| unsafe{
-                            w.$usartX_remap().$bit(($closure)(PINS::REMAP))
+                            w.$usartX_remap().$bit(($closure)(REMAP::REMAP))
```
TheZoq2: This threw me off quite a bit. Suggested comment:

```
// The USARTs have different amounts of remap options which may use 1 or 2 bits in the configuration register. The $bit variable selects wether the size is 1 or 2, and the closure is a closure which returns a bool if 1 bit and a u8 otherwise
```

burrbull:
I think we should decide what variant we prefer to use for "derived" peripherals:

1. Macros everywhere
2. `Deref` where it is possible and `RegisterBlock` where it's not.
3. `RegisterBlock` everywhere

TheZoq2:
I'm not sure what I prefer. To me, the macro version is a bit ugly but more readable.

TheZoq2:
Another argument against using RegisterBlock is that the docs seem to no longer give any information about which actual pins implement the various `Pins` traits. For example https://docs.rs/stm32f1xx-hal/0.5.3/stm32f1xx_hal/pwm_input/trait.Pins.html

burrbull:
Closed in favour of #182

PR #170, opened 2020-01-17
Fix compile error for missing model feature

dbrgn:
The cargo argument is `--features`, not `--feature`.


TheZoq2:
Good catch, thanks!

PR #171, opened 2020-01-17
Impl Read<u8> / Write<u8> for Serial

### LINKED ISSUE #167: Impl Read<u8> / Write<u8> for Serial
Is there a reason why Read<u8> / Write<u8> is only implemented for `Rx` and `Tx` (obtainable via `split()`) but not on the `Serial` type directly?

Other HALs like the [stm32f4xx-hal](https://docs.rs/stm32f4xx-hal/0.6.0/stm32f4xx_hal/serial/struct.Serial.html) have such impls.

I'm working with a driver that wants a `Read<u8> + Write<u8>` type and I don't want to create a wrapper type with my own impl if not necessary :slightly_smiling_face:

dbrgn:
Fixes #167.

I managed to implement this without having split tx/rx pins in the `Serial` struct thanks to the help of @Disasm. Instead, it uses a private trait (`UsartReadWrite`) to avoid having duplicated peripheral access logic. There should be no breaking change in this PR.

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -115,75 +115,6 @@ impl Pins<USART3> for (PD8<Alternate<PushPull>>, PD9<Input<Floating>>) {
     const REMAP: u8 = 0b11;
 }
 
-/// Internal trait for the serial read / write logic.
-trait UsartReadWrite: Deref<Target=crate::stm32::usart1::RegisterBlock> {
-    fn read(&self) -> nb::Result<u8, Error> {
-        let sr = self.sr.read();
-
-        // Check for any errors
-        let err = if sr.pe().bit_is_set() {
-            Some(Error::Parity)
-        } else if sr.fe().bit_is_set() {
-            Some(Error::Framing)
-        } else if sr.ne().bit_is_set() {
-            Some(Error::Noise)
-        } else if sr.ore().bit_is_set() {
-            Some(Error::Overrun)
-        } else {
-            None
-        };
-
-        if let Some(err) = err {
-            // Some error occured. In order to clear that error flag, you have to
-            // do a read from the sr register followed by a read from the dr
-            // register
-            // NOTE(read_volatile) see `write_volatile` below
-            unsafe {
-                ptr::read_volatile(&self.sr as *const _ as *const _);
-                ptr::read_volatile(&self.dr as *const _ as *const _);
-            }
-            Err(nb::Error::Other(err))
-        } else {
-            // Check if a byte is available
-            if sr.rxne().bit_is_set() {
-                // Read the received byte
-                // NOTE(read_volatile) see `write_volatile` below
-                Ok(unsafe {
-                    ptr::read_volatile(&self.dr as *const _ as *const _)
-                })
-            } else {
-                Err(nb::Error::WouldBlock)
-            }
-        }
-    }
-
-    fn write(&self, byte: u8) -> nb::Result<(), Infallible> {
-        let sr = self.sr.read();
-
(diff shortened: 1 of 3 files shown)


TheZoq2:
Seems like a nice and simple change, good job! The last matrix message seems to suggest that you may want to make additional improvements. Is that the case or is it ready for merging?

dbrgn:
> The last matrix message seems to suggest that you may want to make additional improvements. Is that the case or is it ready for merging?

I couldn't figure out how to implement the change suggested by @Disasm, but maybe I misunderstood something...

To me, it seems better to keep the trait methods as associated functions without a &self reference, since they actually *are* static and don't require a reference to an instance.

Disasm:
Passing `&self` ensures ownership, but it is something we do not have in all of the cases anyway. I'm trying to improve this code locally it's dfficult to implement this way. So LGTM. Maybe I will improve this in the future.

Disasm:
Well, I have an update. Should I open a PR to the dbrgn's PR or should I PR my changes here after this PR gets merged?

dbrgn:
> Well, I have an update. Should I open a PR to the dbrgn's PR or should I PR my changes here after this PR gets merged?

If you want, you can also send me the patch (`git show > patch.txt`) and I'll add it to this PR :slightly_smiling_face:

Disasm:
Here we are: https://github.com/dbrgn/stm32f1xx-hal/pull/1

dbrgn:
Thanks! I merged your changes, will add a comment.

src/serial.rs
```diff
@@ -357,112 +420,42 @@ macro_rules! hal {
...
-                    }
-                }
-            }
-
             impl crate::hal::serial::Read<u8> for Rx<$USARTX> {
                 type Error = Error;
 
                 fn read(&mut self) -> nb::Result<u8, Error> {
-                    $USARTX::read()
+                    unsafe { &*$USARTX::ptr() }.read()
```
dbrgn: Out of curiosity, what's the advantage of taking `&self` instead of using associated methods, if the new approach even requires an unsafe raw pointer dereference?

Disasm:
With this implementation only one copy of `UsartReadWrite::{read/write/flush}` is created in debug, as opposed to the original implementation with macros.

src/serial.rs
dbrgn: Ah, the binary gets smaller (as mentioned in a comment [below](https://github.com/stm32-rs/stm32f1xx-hal/pull/171#issuecomment-575720942)). I guess that's a good tradeoff?

Disasm: Originally, this unsafe raw pointer dereference was everywhere. Now we have it only when necessary.

dbrgn: Ahh, you're right, I missed that! Much better indeed :slightly_smiling_face:

dbrgn:
I rebased the branch and squashed my fixup commits. I also integrated the "Strip confusing comment" commit into the commit "Deduplicate USART driver code".

From my side, this PR is now ready.

Disasm:
@TheZoq2 could you review this one more time?

TheZoq2:
Looks great, thank you both for your work!

TheZoq2:
Oh yea, just one minor thing. Would you mind adding a changelog entry before I merge?

dbrgn:
> Oh yea, just one minor thing. Would you mind adding a changelog entry before I merge?

Sure! Forgot about that. Done!

TheZoq2:
Lovely!

PR #173, opened 2020-01-21
Update RTFM-based examples

eupn:
This PR bumps the RTFM version to `0.5.x` and refactors RTFM-based examples to the newer RTFM API.


examples/timer-interrupt-rtfm.rs
```diff
@@ -54,48 +58,48 @@ const APP: () = {
...
         init::LateResources {
-            LED: led,
-            TIMER_HANDLER: timer,
+            led,
+            timer_handler: timer,
         }
     }
 
+    // This is optional and can be removed. When removed, RTFM automatically creates an
+    // idle loop with WFI as its body.
```
Disasm: Maybe this comment should be added instead:
https://github.com/stm32-rs/stm32f1xx-hal/pull/52/files#r270492351

Disasm:
As for the `usb_serial_rtfm` example, change looks good to me.

examples/timer-interrupt-rtfm.rs
eupn: Updated the comment.

examples/timer-interrupt-rtfm.rs
```diff
@@ -54,48 +58,51 @@ const APP: () = {
...
-            // Waits for interrupt
-            wfi();
+            cortex_m::asm::wfi();
         }
     }
 
-    #[interrupt(priority = 1, resources = [LED, TIMER_HANDLER, LED_STATE])]
-    fn TIM1_UP() {
+    #[task(binds = TIM1_UP, priority = 1, resources = [led, timer_handler, led_state])]
+    fn tim1_up(cx: tim1_up::Context) {
```
TeXitoi: You can give a meaningful name to this function, as tick for example.

examples/timer-interrupt-rtfm.rs
```diff
@@ -22,15 +21,20 @@ use stm32f1xx_hal::{
...
+        timer_handler: CountDownTimer<pac::TIM1>,
+
+        #[init(false)]
+        led_state: bool,
+    }
+
     #[init]
-    fn init() -> init::LateResources {
+    fn init(cx: init::Context) -> init::LateResources {
+        let device = cx.device;
```
TeXitoi: You can also declare cx as mut, and then use cx. when needed. Up to you.

examples/usb_serial_rtfm.rs
```diff
@@ -43,7 +47,7 @@ const APP: () = {
         // This forced reset is needed only for development, without it host
         // will not reset your device when you upload new firmware.
         let mut usb_dp = gpioa.pa12.into_push_pull_output(&mut gpioa.crh);
-        usb_dp.set_low();
+        let _ = usb_dp.set_low();
```
TeXitoi: I prefer unwrap here.

TeXitoi:
Thanks for the update!

PR #176, opened 2020-01-21
Implement full embedded_hal::Pwm

justacec:
This change implements the full `embedded_hal::Pwm` and  `embedded_hal::PwmPin` API defined at 

Trait | `embedded_hal` Documentation Link
-----|---------------------------------------------
Pwm | [https://docs.rs/embedded-hal/0.2.3/embedded_hal/trait.Pwm.html](https://docs.rs/embedded-hal/0.2.3/embedded_hal/trait.Pwm.html)
PwmPin | [https://docs.rs/embedded-hal/0.2.3/embedded_hal/trait.PwmPin.html](https://docs.rs/embedded-hal/0.2.3/embedded_hal/trait.PwmPin.html). 

# Motivation
This PR was prompted by the desire to be able to adjust the frequency of the timer after construction through the `set_period()` and `get_period()` `Pwm` trait methods.  The changes were done in such a way as to maintain compatibility with the current API.

# Coding Concept
The primary idea was to shift the current `stm32f1xx_hal::pwm::Pwm` struct to be `PwmChannel` and then devise a new `Pwm` struct which held the channels.  While this sounds potentially not necessary, it was required in order to have a landing space for the `set_period` and `get_period` methods in the `embedded_hal::Pwm` trait.  Implementing this method on an individual channel is not preferred because the operation adjusts the Prescaler (PSC) and Auto Reload Register (ARR) of the entire timer which affects all the channels, not just one.  Implementing the desired method on the `PwmChannel` struct could lead to unintended consequences.

## Time Units
The `embedded_hal` API requires that a time based type be defined for the `Pwm` trait.  This is difficult as the natural timescale for the MCU is in fractional seconds of high order.  If one wanted to use milliseconds (the currently defined time unit in the `stm32fxx_hal` crate) periods greater than 2 seconds could not be specified because of the 32-bit field depth.  If one was to try to use microseconds, there could be difficulty getting particular frequencies because of the integer nature of the unit

Because of this, I have defined the time unit in the `Pwm` trait to be `Hertz`.  I do realize this is not a "time".  This unit is just a bit easier to work with.  In order to make it so end users can pass time like units, I have implemented `Into<Hertz>` for both `MilliSecond` and `MicroSecond`  this allows for adjusting the period like the following:

```
mypwm.set_period(1.ms());
```

# Compatibility
In order to maintain compatibility with the previous API, I have made `Pwm` `Deref` and `DerefMut` in such a way that it returns a reference to the channels.  I have further made `PwmChannel` `Clone` and `Copy`.  This is safe as the `PwmChannel` does not occupy any space as it is a struct holding nothing but `PhantomData` elements.  The struct and its fields carry only type information.

# Random Thoughts

 1. I have been calling the channel in this PR as `PwmChannel`, but the `embedded_hal` API calls it `PwmPin`.  Should we use `PwmPin` here?  I like channel as that is the way that the ST documentation has it.  Also, different channels could point to different pins.  With that thought I might suggest that the `embedded_hal` API be updated/adjusted to match this nomenclature.

# Potential Improvements

While coding this up I recognized a few potential improvements but decided to not implement them right now in order to get a working base.  Below is a list of these ideas which were not implemented:

 ## Array based Channels
Currently, the channels are held in a compile time tuple with length specified by the number of channels in the `PINS` object.  Indexing these channels is then done by their order in the tuple.  This results in cases where the timers `C3` channel is indexed as 0, not 3.  It might be more clear if the `PwmChannel` objects are stored as `Option` in a fixed length array of size 5 (leave the first entry as a `None` so the numbers are more logical when compared with the ST manuals).  That way if `C3` is the only pin defined, it is accessed by:

```
mypwm.3
```

In the event that a user asks for a pin that was not initialized, he/she would get a None and would need to handle that situation.


src/pwm.rs
```diff
@@ -258,10 +370,59 @@ macro_rules! hal {
...
-                unsafe { mem::MaybeUninit::uninit().assume_init() }
+                Pwm {
+                    clk: clk,
+                    _tim: PhantomData,
+                    _channels: PhantomData
+                }
+            }
+            
+            pwm_impl!(
+                $TIMX, $timX, (0, 1, 2, 3), (C1, C2, C3, C4);
```
burrbull: Looks like you make 14 implementation for each combination of `PINS::Channels`. (14*14)
This is incorrect. Only 1 implementation must be here.
If you port this from somewhere, please add link.

justacec: Not quite.  I do 14 combinations for each defined timer from the from `hal!` expansions.  So, in this case it would be 14*4.  The impl is required multiple times because of the concretely typed channel configurations.  I tried to do it outside with generics but it would not work.

justacec: Also, I did not port any code from other sources.  I just tried to mimic the existing coding style in this file.  I did get some assistance in the Discord concerning macros.  But that was far from a copy-paste operation.

justacec: All that being said, if there was a way to do the impl in there without the macro expansion, I agree, less code for the compiler, the better...

burrbull: Sorry, you are right. I thought that this is inside `fn $timX<REMAP, P, PINS>(`.

burrbull:
LGTM.

cc @geomatsi

justacec:
@burrbull There is still an existing issue with a compiler warning on the uninit memory.  I was not sure how to tell the compiler to disreguard that issue.  Do you know how?

justacec:
I am not seeing that warning in the TravisCI build logs.  very strange.

src/pwm.rs
```diff
@@ -209,6 +234,93 @@ pub struct C2;
 pub struct C3;
 pub struct C4;
 
+/*
+  The following implemention of the embedded_hal::Pwm uses Hertz as a time type.  This was choosen 
+  because of the timescales of operationsbeing on the order of nanoseconds and not being able to
```
TheZoq2: operations being

src/pwm.rs
```diff
@@ -209,6 +234,93 @@ pub struct C2;
...
+  could be resolution issues when trying to get a specific value, because of the integer nature.
+
+  To find a middle ground, the Hertz type is used as a base here and the Into trait has been 
+  defined for several base time units.  This will allow for calling the set_period method with 
+  something that is natural to both the MCU and the end user.
+*/
+macro_rules! pwm_impl {
+    ( $( $TIMX:ident, $timX:ident, ( $($CHNUM:tt),+ ), ( $($ENCHX:ident),+ ); )+ )  => {
+            $(
+            #[allow(unused_parens)]
```
TheZoq2: I'm not a fan of this, where is it needed?

src/pwm.rs
```diff
@@ -209,6 +234,93 @@ pub struct C2;
...
+                }
+
+                fn get_period(&self) -> Self::Time {
+                    let clk = self.clk;
+                    let mut psc: u16 = 0;
+                    let mut arr: u16 = 0;
+                    unsafe {
+                        psc = (*$TIMX::ptr()).psc.read().psc().bits();
+                        arr = (*$TIMX::ptr()).arr.read().arr().bits();
+                    }
```
TheZoq2: I prefer 
```rust
let mut psc = unsafe{*$TIMX::...};
```

burrbull:
This is because you've added `clk: Hertz` to `Pwm`.
You can just use `#[derive(Clone, Copy)]` on `Pwm`  instead of implement `Clone` and `Copy` manually.

But I don't understand sense `Deref`.

src/pwm.rs
```diff
@@ -258,10 +370,59 @@ macro_rules! hal {
...
+                fn clone(&self) -> Self {
+                    unsafe { mem::MaybeUninit::uninit().assume_init() }
+                }            
+            }
+
+            impl Copy for PwmChannel<$TIMX, C4> {}
+            impl Clone for PwmChannel<$TIMX, C4> {
+                fn clone(&self) -> Self {
+                    unsafe { mem::MaybeUninit::uninit().assume_init() }
+                }
```
TheZoq2: I'm pretty sure that this is unsound. By cloning a channel, you can pass the same channel to two different places at which point `set_duty` would affect both clones.

justacec:
@burrbull, The Clone / Copy is needed against the `PwmChannel`.  I did try to put the derive directive above the `PwmChannel` struct definition, but that did not seem to work and just produced errors.  It was not until I manually implemented Clone that everything worked.  

What is your specific question concerning Deref?

src/pwm.rs
justacec: "Operations" in this context was intended to be the bus clock period.  Maybe the text here needs to be cleaned up a bit.  I was trying to help anybody who was reading this understand why the units of my time type is inverse time.

src/pwm.rs
justacec: Are you meaning the `#[allow(unused_parens)]`?  It might not need to be here.  This came from the macro defined at the top of the file.  I agree, if it does not need to be here, lets get rid of it.  I will look at it tonight.

src/pwm.rs
justacec: Ohh, I like that better as well.  Again, will do this tonight.

src/pwm.rs
justacec: This is true.  But, this is a hardware device and the channel maps directly to a physical device.  If an end user clones it he or she must know that if the hardware changes it will affect the software processes referring to those devices.

Ok, with that out of the way, do you know how to do this such that there is ownership?  I would be more than happy to change this.

TeXitoi: I think you want `Pwm::into_channels(self) -> PWMCHANNELS`. That would allow to separate the channels and then use them independently.

src/pwm.rs
```diff
@@ -258,10 +370,59 @@ macro_rules! hal {
...
+            
+            pwm_impl!(
+                $TIMX, $timX, (0, 1, 2, 3), (C1, C2, C3, C4);
+                $TIMX, $timX, (0, 1, 2), (C1, C2, C3);
+                $TIMX, $timX, (0, 1, 2), (C1, C2, C4);
+                $TIMX, $timX, (0, 1, 2), (C2, C3, C4);
+                $TIMX, $timX, (0, 1), (C1, C2);
+                $TIMX, $timX, (0, 1), (C1, C3);
+                $TIMX, $timX, (0, 1), (C1, C4);
+                $TIMX, $timX, (0, 1), (C2, C3);
```
TeXitoi: That's quite disturbing. I'd prefer an `pub enum Channel { C1, C2, C3, C4 }` as `hal::Pwm::Channel` and then just having `Channel::C2` to manage `C2`

src/pwm.rs
TheZoq2: Oh sorry, I guess I wasn't clear here. It currently says "operationsbeing", but should be "operations being" with a space :)

src/pwm.rs
TheZoq2: >  If an end user clones it he or she must know that if the hardware changes it will affect the software processes referring to those devices.

This goes against the idea of this crate and most of the rust embedded ecosystem. The idea is to use rusts ownership model to make sure that there is never any action at a distance.

An example of situations that this prevents (pseudocode which doesn't really use this API, it's late so I can't be bothered to make i correct. It should demonstrate the issue though):

```rust
let pwm = // pwm...;
let (c1, c2) = pwm.channels();

// Some independent driver that uses c1 and relies on a specific period
run_driver(c1, ...);

// Some other driver that uses the Pwm trait to change the period, thus affecting both c1 and c2
run_driver2(c2, pwm);

// By adding the second driver, we broke the behaviour of the first which will be a pain to debug, and debugging it would require knowledge of both driver1, driver2 and the interraction between them.
```

> Ok, with that out of the way, do you know how to do this such that there is ownership? I would be more than happy to change this.

I could be wrong about this, but I presume the reason you need copy is that otherwise you can't call `set_period` on the pwm struct after distributing the channels elsewhere.

If that is the case, I can't think of a super neat way to do things off the top of my head. I still think the thing we discussed in discord where in order to set the period, you need to guarantee that the pins are not being used elsewhere (you can give mutable references) is the best option.

Unfortunately, that makes the API a lot less inconvenient, and I suspect that implementing everything in `Pwm` might be difficult.

Hopefully my sleepy ramblings provided some insight. If something is unclear, I'll hopefully be able to give a better answer tomorrow when it's less late ;)

examples/pwm.rs
```diff
@@ -49,25 +50,57 @@ fn main() -> ! {
...
 
     let mut pwm = Timer::tim2(p.TIM2, &clocks, &mut rcc.apb1)
-        .pwm::<Tim2NoRemap, _, _, _>(pins, &mut afio.mapr, 1.khz())
-        .2;
+        .pwm::<Tim2NoRemap, _, _, _>(pins, &mut afio.mapr, 1.khz());
+
+    //// Operations affecting all defined channels on the Timer
+
+    // Adjust period to 0.5 seconds
+    pwm.set_period(500.us());
```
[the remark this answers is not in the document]
justacec: Ok, I think I got that fixed.

src/pwm.rs
```diff
@@ -199,7 +203,28 @@ impl Timer<TIM4> {
...
+    clk: Hertz,
+    _channels: PhantomData<PWMCHANNELS>,
+    _tim: PhantomData<TIM>,
+}
+
+impl<TIM, PWMCHANNELS> Deref for Pwm<TIM, PWMCHANNELS> {
+    type Target = PWMCHANNELS;
+
+    fn deref(&self) -> &Self::Target {
+        unsafe { mem::MaybeUninit::uninit().assume_init() }
```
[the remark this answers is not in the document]
justacec: Made change

src/pwm.rs
```diff
@@ -199,7 +203,28 @@ impl Timer<TIM4> {
     }
 }
 
-pub struct Pwm<TIM, CHANNEL> {
+pub struct Pwm<TIM, PWMCHANNELS> {
+    clk: Hertz,
+    _channels: PhantomData<PWMCHANNELS>,
+    _tim: PhantomData<TIM>,
```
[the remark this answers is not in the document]
justacec: Done

src/pwm.rs
```diff
@@ -199,7 +203,28 @@ impl Timer<TIM4> {
...
+    }
+}
+
+impl<TIM, PWMCHANNELS> DerefMut for Pwm<TIM, PWMCHANNELS> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        unsafe { mem::MaybeUninit::uninit().assume_init() }
+    }
+}
+
+#[derive(Copy, Clone)]
```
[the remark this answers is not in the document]
justacec: I was needing that in order to mimic the existing API.  Without it I cannot do `let a = mypwm.0`. Lets talk about this.

src/pwm.rs
```diff
@@ -199,7 +203,28 @@ impl Timer<TIM4> {
...
+    type Target = PWMCHANNELS;
+
+    fn deref(&self) -> &Self::Target {
+        unsafe { mem::MaybeUninit::uninit().assume_init() }
+    }
+}
+
+impl<TIM, PWMCHANNELS> DerefMut for Pwm<TIM, PWMCHANNELS> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        unsafe { mem::MaybeUninit::uninit().assume_init() }
```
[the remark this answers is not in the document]
justacec: Completed

src/pwm.rs
justacec: Corrected

src/pwm.rs
justacec: Removed

src/pwm.rs
justacec: Corrected

src/pwm.rs
burrbull: > `Pwm::into_channels(self) -> PWMCHANNELS`

or `split()`

src/pwm.rs
justacec: I will implement without the Clone/Copy and use the `into_channels(&self)` approach.  I was just trying to maintain the current API.  I also like the Enum approach as well.  I will look to try and implement that as well.

burrbull:
@justacec  See this draft https://github.com/burrbull/stm32f1xx-hal/commit/42acde3193329b97eee3b13c39dd5cb73413f282

justacec:
So I feel like there are some conflicting suggestions from different reviewers.  The one that is bothering me right now is whether to keep the `PhantomData` in the `Pwm` struct?  Does everybody agree that should go away?

Lots of moving parts right now.  I am going to hide in my bed for right now.  :)

src/pwm.rs
TeXitoi: That's really `into_channels(self)` without `&`

TheZoq2:
If the `PhantomData` is not required, I'd say it should be removed

jounathaen:
I'm not particularly happy with using `Hertz` as argument for Period.

Just some random thoughts:
Would it be possible to add a seconds/milliseconds type to `stm32f1xx_hal::time` which can be converted to `stm32f1xx_hal::time::*Hertz` and vice versa?
(There must be some other needs for a seconds/milliseconds/... type appart from PWM. For example timers.)

In addition I'd suggest to add the complementary function for the pwm `pwm::set_frequency()` (both can of course share the implementation).

This could be proposed as a addition to the generic embedded-hal.

justacec:
@jounathaen I had your same issues but after looking at the time scales came to the conclusion, along with @TheZoq2, that Hertz was the right thing to use.  This is because of the time timescales involved and the limitation of staying in only 32-bit integer space.  See my commentary above.  

As far as the `set_frequency` methods, sure.  That sounds like a good idea.  I was just trying to be strict for the `embedded_hal` traits which do not define that.

justacec:
@burrbull I have looked at your alternative and it seems to be going in the right direction for what you were looking for.  Right now I am tempted to just let you finish this out.  It looks like I have done what needed to be done to spur this API change.

I thought about just pulling over your code into my branch but then that just seemed like cheating.

TheZoq2:
> Would it be possible to add a seconds/milliseconds type to stm32f1xx_hal::time which can be converted to stm32f1xx_hal::time::*Hertz and vice versa?

This is something I would like to do as well. As @justacec says, it has the limitation that you can accidentally specify an out of bounds value. Perhaps a runtime check is fine here and in the timers. Unfortunately I don't think the embedded_hal traits allow returning a `Result` from the relevant functions.

In general, I'm not very happy with how time works throughout the `embedded_hal` ecosystem (at least the parts I've touched). My main gripe with it is that there is no way for a driver to specify requirements about time units, but that's probably a discussion for another day.

> I thought about just pulling over your code into my branch but then that just seemed like cheating.

Cheat as much as you like if it leads to better end result :) I don't mind which version we use, work with the one that is the furthest along.

If you want to improve the code in that PR, you can also suggest changes in the github interface or comments. You can also do a PR to the source branch of the PR if you want to suggest more complex changes.

burrbull:
Use my code like you want. Anyway it was your idea.

jounathaen:
> @jounathaen I had your same issues but after looking at the time scales came to the conclusion, along with @TheZoq2, that Hertz was the right thing to use. This is because of the time timescales involved and the limitation of staying in only 32-bit integer space. See my commentary above.

Sure, but then I'd suggest to rename the function `set_frequency` and NOT implementing `set_period(Hertz)`, until there is a better concept of time available. Of course, in that case the PWM trait still is not implemented, but imagine the hazzle, if the API has to be changed in the future.
(I'd rather not implement the trait, than introducing a suboptimal API.)

TeXitoi:
@jounathaen it doesn't surprise me as the difference between a frequence and a period is the unit, and that's codded in the type system, so no error possible.

We can change the associated type later, that's not as if we do breaking changes several times by years.

therealprof:
> So I feel like there are some conflicting suggestions from different reviewers.

@justacec It's fantastic that you decided to tackle PWM. The many comments are proof for the vast interest in this. I'm quite confident @TheZoq2 and @TeXitoi will help you land this in a useful manner, so I will refrain from engaging in the technical discussion.

Looks great so far, thanks a lot!

examples/pwm.rs
```diff
@@ -49,25 +51,56 @@ fn main() -> ! {
...
+    asm::bkpt();
 
     let max = pwm.get_max_duty();
 
-    pwm.enable();
+    //// Operations affecting single channels can be accessed through
+    //// the Pwm object or via dereferencing to the pin.
+
+    // Use the Pwm object to set C3 to full strength
+    pwm.set_duty(Channel::C2, max);
```
burrbull: You wrote `C3` but use `C2`

justacec: Opps.  Corrected this.

justacec:
@burrbull @TheZoq2 @TeXitoi Ok, I believe that I have merged all of the efforts in the other pull request (#177) to this pull request.  I think that I might have some concerns about how the channels are dereferenced by the `split` result.  I will expand on this a bit later.

src/pwm.rs
justacec: @TeXitoi I think this comment is OBE based on the updated code I pulled from @burrbull. Please reopen if not.

src/pwm.rs
justacec: @TeXitoi I am going to close this "sub-conversation" and ask for a re-review because of the large changes that were put in by @burbull

justacec:
Ok, so, what I was getting to earlier about the `split` issue is that you could have a custom channel set with a single pin which is, by the ST documentation bound to Channel 2 of the timer.  While that channel would be referenced as `Channel::C3` it would be dereferenced as `split().0` and not `split.3`.  What do you guys think?  I think that I am just looking for some consistency.

justacec:
@burrbull For some reason, I noticed that I cannot specify that I want you to re-review the code.  Not sure why.  That being said, can you please take a look.  Thanks.

src/pwm.rs
```diff
@@ -80,6 +82,26 @@ pub trait Pins<REMAP, P> {
...
+
+    fn check_used(c: Channel) -> Channel {
+        if (c == Channel::C1 && Self::C1)
+            || (c == Channel::C2 && Self::C2)
+            || (c == Channel::C3 && Self::C3)
+            || (c == Channel::C4 && Self::C4)
+        {
+            c
+        } else {
+            panic!("Unused channel")
```
TheZoq2: Still not a fan of this panic, and I'm not sure if check_used is even nessecary anymore

src/pwm.rs
```diff
@@ -54,6 +54,7 @@
 */
 
 use core::marker::PhantomData;
+use core::marker::{Copy};
```
TheZoq2: This is unused now, right?

justacec: @TheZoq2 Copy is used for:

```
#[derive(Clone, Copy, PartialEq)]
pub enum Channel {
    C1,
    C2,
    C3,
    C4,
}
```

src/pwm.rs
justacec: @TheZoq2 It took me a bit to understand as this was a @burbull construction, but the function seems to make sense to me.  Without the function there is no way for the per-channel functions to know if that channel is, in-fact, being used.  As an example, lets say that I configure the Timer to use only Channel 3 passing a single pin.  But then I call something like:

```
mypwn.enable(Channel::C1);
```

Without the `check_used` method, the function would go ahead and set the bit.  *Although, I am not sure that are any real negative consequences of this.*

This was the real reason behind me using a macro in my original incarnation of this functionality.  The macro expanded to only include channels that were active and would default to doing nothing through a default match.  In this incantation, this has been shifted to be run-time vice compile-time.

Now, concerning the panic.  Although I had originally default to no action, I actually see no issue with this panic as it will aggressively inform the user that he/she has something amis.  I kind of feel that this is a better approach than doing nothing.  In the do nothing variant, I could mis-type something and work for hours trying to figure out why my channel duty is not acting correctly; only to find out that I had put the wrong channel in there.  If there was a panic, the compiler would yell at me (I think). If I am mis-understanding that it would occur at the compiler phase and it would be runtime, I would have a potential backtrace which would point to the exact place that needed to be checked.

TheZoq2: > Now, concerning the panic. Although I had originally default to no action, I actually see no issue with this panic as it will aggressively inform the user that he/she has something amis.

Agreed, panic is better than doing nothing

>  If there was a panic, the compiler would yell at me (I think). If I am mis-understanding that it would occur at the compiler phase and it would be runtime,
Panics happen at runtime, if possible I would like to move this to compile time but that is a difficult task.

> If I am mis-understanding that it would occur at the compiler phase and it would be runtime, I would have a potential backtrace which would point to the exact place that needed to be checked.

Yes, but backtraces can be a bit of a pain to get a hold of in embedded contexts

But again, we can probably take care of this in the future in another PR, so unless burbull has additional comments, I think this is good to go

burrbull: I don't know how to do this check compile time because `channel` can be not constant.

In fact we need possibility to write `const fn` in trait implementation and variant of `assert` which fails in compile when it is possible and fail in runtime in other cases.

burrbull:
> Ok, so, what I was getting to earlier about the `split` issue is that you could have a custom channel set with a single pin which is, by the ST documentation bound to Channel 2 of the timer. While that channel would be referenced as `Channel::C3` it would be dereferenced as `split().0` and not `split().3`. What do you guys think? I think that I am just looking for some consistency.

Alternative for this is only one: to use 4 pins and 4 channels everywhere and to use `NoPin` (or `NoChannel`) for absent pins. This even easier to implement.

justacec:
@burrbull In that case it would always return a size 4 tuple then right.

justacec:
@burrbull So, I have now been looking into trying to implement this.  Right now I am trying to find out how to adjust the existing `pins_impl` macro such that the `Channels` type is always a size 5 tuple with the channels in the right place.  Right now it looks like that would need to be adjusted to be a procedural macro to have the flexibility to format the `Channels` type.  While I look into learning `proc_macro`, did you have any other thoughts on how this could be implemented that might steer me in the right direction.  

BTW: I am returning a size 5 tuple so that channel 1 will be dereferenced as `mychannels.1` and not `mychannels.0`.

Thanks for any direction.

TeXitoi:
Please no proc macro and no 5 element tuple. If you want logical fields, better to create a struct with chan1, chan2... fields.

burrbull:
1. I can suggest to add `NoChannel` (or `NoPin`) fictive `struct` like in SpI/I2C.
https://github.com/burrbull/stm32f1xx-hal/commit/1f4b04a169fbd8f7dac42991a2adf20092bd3d27

2. `Pwm` and `PwmChannel` can own by tuple of 4 real pin structs instead of phantom `PINS`.
    That allows us to release timer and pins.

3. And yes, 5 element tuple is not good idea.

justacec:
@burrbull @TheZoq2 

Guys,  

Sorry I have been absent the past several weeks.  My work has really stepped up and I have not been able to dedicate as much time to trying to work on my personal projects, which includes learning Rust.  I did attempt to get something working which would return a full object vice the Tuple to get around the issue that I had raised earlier.  I think that I was very close but, not quite there.  I think that you guys would probably be able to implement a fix for that pretty quickly.  

I would like to propose that we go ahead and merge this pull request in it's current state as it maintains the current API and provides a fuller implementation of the `embedded_hal` API.  Once this is in, somebody, possibly even me again (with some coaching :) ) could tackle the tuple return issue.  

Thoughts?

PR #177, opened 2020-01-22
Alternative for #176 

burrbull:


src/pwm.rs
```diff
@@ -258,10 +300,95 @@ macro_rules! hal {
...
+                    }
+                }
+
+                fn get_duty(&self, channel: Self::Channel) -> Self::Duty {
+                    match PINS::check_enabled(channel) {
+                        Some(Channel::C1) => unsafe { (*$TIMX::ptr()).ccr1.read().ccr().bits() },
+                        Some(Channel::C2) => unsafe { (*$TIMX::ptr()).ccr2.read().ccr().bits() },
+                        Some(Channel::C3) => unsafe { (*$TIMX::ptr()).ccr3.read().ccr().bits() },
+                        Some(Channel::C4) => unsafe { (*$TIMX::ptr()).ccr4.read().ccr().bits() },
+                        None => 0,
```
TheZoq2: Is 0 always equivalent to a disabled channel? Or can the channel be in some sort of floating state if disabled?

src/pwm.rs
```diff
@@ -258,10 +300,95 @@ macro_rules! hal {
...
+                    }
+                }
+
+                fn set_duty(&mut self, channel: Self::Channel, duty: Self::Duty) {
+                    match PINS::check_enabled(channel) {
+                        Some(Channel::C1) => unsafe { (*$TIMX::ptr()).ccr1.write(|w| w.ccr().bits(duty)) },
+                        Some(Channel::C2) => unsafe { (*$TIMX::ptr()).ccr2.write(|w| w.ccr().bits(duty)) },
+                        Some(Channel::C3) => unsafe { (*$TIMX::ptr()).ccr3.write(|w| w.ccr().bits(duty)) },
+                        Some(Channel::C4) => unsafe { (*$TIMX::ptr()).ccr4.write(|w| w.ccr().bits(duty)) },
+                        None => {},
```
TheZoq2: In this code:
```rust
pwm.disable(C1);
pwm.set_duty(C1, 0.5);
pwm.enable(C1);
```
I would expect it to get the updated duty after being enabled, but this seems to not be the case here. As far as I can tell, 

```rust
match channel {
    Channel::C1 => unsafe{ /*ccr1.write...*/ }
    Channel::C2 => unsafe{ /*ccr2.write...*/ }
    Channel::C3 => unsafe{ /*ccr3.write...*/ }
    Channel::C4 => unsafe{ /*ccr3.write...*/ }
}
```
would result in that behaviour

src/pwm.rs
```diff
@@ -258,10 +300,95 @@ macro_rules! hal {
...
+                    let mut psc: u16 = 0;
+                    let mut arr: u16 = 0;
+                    unsafe {
+                        psc = (*$TIMX::ptr()).psc.read().psc().bits();
+                        arr = (*$TIMX::ptr()).arr.read().arr().bits();
+                    }
+
+                    // Length in ms of an internal clock pulse
+                    (clk.0 / u32(psc * arr)).hz()
+//                    (((psc as u32) / clk.0) * (1_000_000 as u32) * (arr as u32)).ms()
```
TheZoq2: This should probably be removed

src/pwm.rs
```diff
@@ -258,10 +300,95 @@ macro_rules! hal {
...
+                }
+
+                fn set_period<T>(&mut self, period: T) where
+                    T: Into<Self::Time> {
+                        let clk = self.clk;
+
+//                        let freq = u16(1 / period.into());
```
TheZoq2: This should be removed as well

src/pwm.rs
burrbull: What about to panic here?

burrbull:
I've renamed function to check_used and added panic.
Now you cannot `enable`/`disable` channel if it is not used (pin not specified).

src/pwm.rs
TheZoq2: I would really like to avoid panics as far as possible. I don't see a great way to accomplish that though :/

TeXitoi: By experience, disable is different to duty of 0

TeXitoi: But it should be OK to give the duty whatever it is enabled or disabled, according to enabling the channel would give you this duty.

src/pwm.rs
```diff
@@ -80,6 +81,26 @@ pub trait Pins<REMAP, P> {
...
+    fn check_used(c: Channel) -> Channel {
+        if (c == Channel::C1 && Self::C1)
+            || (c == Channel::C2 && Self::C2)
+            || (c == Channel::C3 && Self::C3)
+            || (c == Channel::C4 && Self::C4)
+        {
+            c
+        } else {
+            panic!("Unused channel")
+        }
```
burrbull: @TheZoq2 @TeXitoi 
In fact this code must be optimized by compiler. So no panic will be in code if user specify right channel.

TheZoq2: Ideally I want to avoid panics if the user specifies the wrong channel as well :) But I don't see a great way to do that at the moment

src/pwm.rs
TheZoq2: Yep, that seems like the best solution

justacec:
I just integrated this into my branch

burrbull:
> I just integrated this into my branch

Ok. I close PR, but branch will be available.

PR #182, opened 2020-02-02
Spi Deref

burrbull:
Part of #161 

Spi only and `Deref`.

`rustfmt` for `spi.rs`

Make `RccBus` sealed instead of `Reset` and `Enable`.

cc @TheZoq2


TheZoq2:
Sorry for the delay!

Looks good apart from the usual question: Would you mind adding a changelog entry? :)

burrbull:
Done.

PR #185, opened 2020-02-16
bump stm32f1 0.10.0

burrbull:


examples/exti.rs
```diff
@@ -61,8 +60,7 @@ fn main() -> ! {
         int_pin.enable_interrupt(&p.EXTI);
     } // initialization ends here
 
-    let mut nvic = cp.NVIC;
-    nvic.enable(pac::Interrupt::EXTI9_5);
+    unsafe { pac::NVIC::unmask(pac::Interrupt::EXTI9_5); }
```
TeXitoi: More unsafe? :-(

burrbull: https://docs.rs/cortex-m/0.6.2/cortex_m/peripheral/struct.NVIC.html#method.enable

deprecated

TeXitoi: OK, thanks for the link.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Bump `stm32f1` dependency (`0.10.0`)
```
therealprof: It's a breaking change, right?

burrbull: I don't remember any change for `f1` in `stm32-rs`.

TeXitoi: Technically it is as this is a semver incompatible version export. If a user depend on stm32f1 0.9 and this half, it will break compilation as the 2 types will be different.

Issue #165, opened 2020-01-13
PWM on advanced timers requires TIMx_BDTR.MOE set to work


cs2dsb:
The advanced timers have a master output enable field in their break & dead time register that must be set before PWM will work.

```
  //This makes the PWM on TIM1 work on my hardware:
  unsafe {
      &(*TIM1::ptr()).bdtr.modify(|_, w| w.moe().set_bit());
  }
```

I'll submit a PR at some point but thought I'd raise the issue in case I forget / someone gets there before me.

PR #187, opened 2020-02-17
Fixes PWM on TIM1

thalesfragoso:
Closes https://github.com/stm32-rs/stm32f1xx-hal/issues/165

Also, now we update the `psc` before starting the timers.


TheZoq2:
Looks great, thanks!

PR #188, opened 2020-02-19
examples/blinky_timer_irq.rs: added an '#[interrupt]' blinky demo

spicyjack:
I saw the `blinky_timer_irq.rs` demo that @therealprof wrote in the `stm32f0xx-hal` repo and wanted to poke at it, but I only have a Black Pill (STM32F103) board on hand, so...

I ran this command on the example file:

    rustfmt --config imports_layout=HorizontalVertical examples/blinky_timer_irq.rs

and I was almost happy with the results, but not quite.

Obligatory animated GIF, since blinkenleds are so exciting...
blinky_timer_irq


spicyjack:
@jamwaffles I've made the capitalization changes

PR #190, opened 2020-03-07
spi-dma example: swap buffer and SPI device in return value of a transfer's wait

trnila:
`wait` method returns tuple of `(BUFFER, TxDma<PAYLOAD, $CX>)`


TheZoq2:
Good catch, thanks for the fix!

PR #198, opened 2020-04-09
Examples: Add timed one shot interrupts

mwkroening:
This adds examples for firing one shot interrupts repeatedly with `RTC` and `TIM2`.

I wrote those to understand this topic better and think it could be useful to others.

What do you think?

### DIFF
--- a/examples/interrupt-rtc.rs
+++ b/examples/interrupt-rtc.rs
@@ -1,0 +1,56 @@
+//! Demonstrates real time clock based one shot interrupts
+
+#![no_std]
+#![no_main]
+
+use cortex_m::asm;
+use cortex_m_rt::entry;
+use cortex_m_semihosting::hprintln;
+use panic_semihosting as _; // panic handler
+use stm32f1xx_hal::{
+    pac::{interrupt, Interrupt, Peripherals, NVIC},
+    prelude::*,
+    rtc::Rtc,
+};
+
+#[interrupt]
+fn RTC() {
+    hprintln!("RTC fired.").ok();
+
+    // Disable the RTC interrupt so this won't fire again
+    NVIC::mask(Interrupt::RTC);
+}
+
+#[entry]
+fn main() -> ! {
+    hprintln!("Entered main.").ok();
+
+    // Extract needed peripherals
+    let dp = Peripherals::take().expect("Peripherals have been taken before");
+    let rcc = dp.RCC.constrain();
+    let mut apb1 = rcc.apb1;
+    let mut pwr = dp.PWR;
+    let mut backup_domain = rcc.bkp.constrain(dp.BKP, &mut apb1, &mut pwr);
+
+    // Create and configure Rtc
+    let mut rtc = Rtc::rtc(dp.RTC, &mut backup_domain);
+    rtc.select_frequency(1.hz());
+    rtc.listen_alarm();
+
+    hprintln!("Entering main loop.").ok();
+    loop {
+        rtc.set_alarm(rtc.current_time() + 1);
+
+        // Clear the RTC interrupt's pending state
+        NVIC::unpend(Interrupt::RTC);
+        unsafe {
+            // Enable the RTC interrupt (again)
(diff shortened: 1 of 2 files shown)


thalesfragoso:
I don't think you should use the `NVIC::mask` in this case, you probably want to clear the interrupt in the peripheral itself, both timer and rtc have methods for that. In that case you will need a `Mutex` to be able to access the timer/rtc in the interrupt context.

Also if you were dealing with multiple interrupts this wouldn't work since you wouldn't know which interrupt to `unmask` in your main loop.

mwkroening:
Thanks for your quick reply!

Your approach is surely generally better suited for a beginner's example, as wrapping the timer peripheral in a mutex and then working on that peripheral in the interrupt handler arguably represents the programmer's intent better, and we can profit from Rust's compile time checker to spot any issues should one change portions of that code.

On the other hand having to enter an interrupt free section each time we want to stop the interrupt from firing again is not zero cost (although I did not measure how much it costs). So I'd argue that if I only need to configure a one shot interrupt of one specific peripheral in one specific location, it could make sense to avoid that interrupt free section.

My use case is porting [ferrous-systems/async-on-embedded/nrf52/timer.rs](https://github.com/ferrous-systems/async-on-embedded/blob/master/nrf52/src/timer.rs) to the Blue Pill. Once familiar with that code one can verify, that it is working correctly. Everything regarding the interrupt is near the interrupt. On the other hand I see the pitfall of carelessly changing that code and then missing verifying these conditions manually.

The compile time checker and general expressiveness are two of rust's strengths but since this is not zero cost, I am currently sitting on the fence on what paradigm to favor there.
What do you think? Would you choose to omit sharing the peripheral?

Regarding the examples: Given that the examples for `Rtc` and `Timer<TIM2>` are fairly similar, I would suggest removing one of them (say `TIM2`) and providing the other one in both variants, one as is and one explicitly sharing the peripheral.

thalesfragoso:
@mwkroening 

>On the other hand having to enter an interrupt free section each time we want to stop the interrupt from firing again is not zero cost (although I did not measure how much it costs). So I'd argue that if I only need to configure a one shot interrupt of one specific peripheral in one specific location, it could make sense to avoid that interrupt free section.

The problem with your example is that it's very contrived, using the `NVIC` that way if you add another interrupt you wouldn't be able to know each one fired in the main loop. If you don't want to use a `Mutex` that disables interrupts you can put the peripheral in a static and **unsafely** access it on the interrupt, that would be okay if you don't access it anywhere else anymore.

You can see something in these lines (but better) here:
https://github.com/stm32-rs/stm32f1xx-hal/blob/master/examples/blinky_timer_irq.rs

Also, I think these examples should be foolproof and use best practices. They're a usually directed at beginners, so it's better to have some extra checking even if that incurs some overhead.

mwkroening:
Ah, I see! In the blinky timer IRQ example we move the peripherals into the interrupt handler via mutexes on the first run, saving time in successive runs. That's also what I started off with my examples.

I need to configure the peripheral outside in the main loop though, so I can only resort to using `unsafe` as you pointed out or `NVIC`, if I don't want the runtime penalty of Mutexes in each run of the interrupt handler.

I still don't quite grasp why using `NVIC` in such a case is bad, sorry. In my case the main loop does not need to know which one of several interrupts fired (the interrupt handlers tell the main loop via `Waker`'s but that might be off-topic). The main loop only needs to prepare such one time interrupts and the interrupt handlers only need to ensure that they are not being run again before being prepared by the main loop again.

Perhaps it would be useful to document what to use the `NVIC` for and what not more clearly.

> Also, I think these examples should be foolproof and use best practices. They're a usually directed at beginners, so it's better to have some extra checking even if that incurs some overhead.

I agree. The examples of this PR might be redundant then, as the Blinky Timer IRQ example seems to be quite complete in this regard, right?

TheZoq2:
> I agree. The examples of this PR might be redundant then, as the Blinky Timer IRQ example seems to be quite complete in this regard, right?

With this in mind, I'll go ahead and close this

PR #202, opened 2020-04-19
initial CRC support

yjh0502:
I'm complete newcomer on microcontroller & stm32, so please feel free to comment.

 - Tested on `stm32f103c8`
 - I'm not sure how it interacts with other peripherals. For example it seems that SPI protocol requires CRC calculation. Does stm32 SPI uses CRC peripherals for CRC calculation of there's another CRC hardware dedicated to SPI CRC?
 - I'm not sure if I should use `split` or `constrain`.
 - small additional diff on `prelude` caused by `rustfmt`.

### DIFF
--- a/src/crc.rs
+++ b/src/crc.rs
@@ -1,0 +1,39 @@
+//! CRC
+
+use crate::pac::CRC;
+use crate::rcc::{Enable, AHB};
+
+/// Extension trait to constrain the CRC peripheral
+pub trait CrcExt {
+    /// Constrains the CRC peripheral to play nicely with the other abstractions
+    fn new(self, ahb: &mut AHB) -> Crc;
+}
+
+impl CrcExt for CRC {
+    fn new(self, ahb: &mut AHB) -> Crc {
+        CRC::enable(ahb);
+        Crc { crc: self }
+    }
+}
+
+/// Constrained CRC peripheral
+pub struct Crc {
+    crc: CRC,
+}
+
+impl Crc {
+    pub fn read(&self) -> u32 {
+        self.crc.dr.read().bits()
+    }
+
+    pub fn write(&mut self, val: u32) {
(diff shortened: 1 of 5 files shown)


src/crc.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub dr: DR,
+    pub cr: CR,
+}
+
+/// Opaque DR register
+pub struct DR {
+    _0: (),
+}
+
+#[allow(dead_code)]
```
TheZoq2, line 33: This probably shouldn't be needed

burrbull: Really, why not just:
```
    let mut crc = Crc::new(p.CRC, &mut rcc.ahb);

    crc.reset();
    crc.write(0x12345678);

    let val = crc.read();
```

yjh0502: Fixed, could you review it again?

src/crc.rs
```diff
@@ -0,0 +1,39 @@
...
+    crc: CRC,
+}
+
+impl Crc {
+    pub fn read(&self) -> u32 {
+        self.crc.dr.read().bits()
+    }
+
+    pub fn write(&mut self, val: u32) {
+        unsafe { self.crc.dr.write(|w| w.bits(val)) }
```
burrbull, line 30: This `unsafe` is unneeded:
https://docs.rs/stm32f1/0.11.0/stm32f1/stm32f103/crc/dr/struct.DR_W.html

yjh0502: Fixed, thanks!

TheZoq2:
Looks good to me, thanks!

TheZoq2:
Actually, would you mind adding an entry to the `Changelog.md` before we merge?

yjh0502:
@TheZoq2 I updated `CHANGELOG.md` :)

TheZoq2:
Lovely, thanks!

PR #203, opened 2020-04-20
Fix race condition causing ADCs to read stale values

### LINKED ISSUE #201: ADC reads are out of order at certain frequencies
At some ADC clock values, the ADC reads values out of order, or to the wrong channels.

```rust
let clocks = rcc.cfgr.freeze(&mut flash.acr);
hprintln!("adc freq: {}", clocks.adcclk().0).unwrap();

let mut adc1 = adc::Adc::adc1(p.ADC1, &mut rcc.apb2, clocks);

let mut ch0 = gpioa.pa0.into_analog(&mut gpioa.crl);
let mut ch1 = gpioa.pa1.into_analog(&mut gpioa.crl);
let mut ch2 = gpioa.pa2.into_analog(&mut gpioa.crl);

loop {
    let r1: u16 = adc1.read(&mut ch0).unwrap();
    let r2: u16 = adc1.read(&mut ch1).unwrap();
    let r3: u16 = adc1.read(&mut ch2).unwrap();

    hprintln!("{}, {}, {}", r1, r2, r3);
}
```

Using the above modified `examples/adc.rs`, with 3.3V connected to to PA0, 0V
connected to PA1, and 2V connected to PA2, the readings returned are
```
2514, 4095, 0
2516, 4095, 0
2519, 4092, 0
2514, 4094, 0
2528, 4094, 0
2531, 4095, 1
2512, 4093, 0
2506, 4095, 0
```
not 4096, 0, 2510 as expected.

This can be fixed by setting the adcclk to 2 MHz in the clock initialisation.

I'm going to work on a patch to fix this, but i'm reporting it now so noone
else has to debug the issue :)

TheZoq2:
Fixes #201

This was a *fun* bug to track down.

Some decided that it would be a good idea to make one way of starting the ADC conversion to write a `1` to the `adon` bit if the `adon` bit is already one. But that might trigger a read when you want to modify other parts of the register, so this behaviour only happens if no other bits are modified. Thus, `modify` on the `cr2` register will start a conversion unless at least one bit is *actually* modified.

This was accidentally done as the first part of the `convert` function, which started a conversion which would be ready roughly by the time the CPU starts checking for `eoc` bits.

The race condition occurs if the CPU gets to the `EOC` check, before the ADC has time to reset the bit as a result of the `swstart` trigger. This relies on the CPU being faster than the ADC clock which explains the behaviour at high prescalers.

Finally, it seems like the line
```rust
while self.rb.cr2.read().swstart().bit_is_set() {}
```
*should* have prevented this as the ADC should have taken care of the `swstart` and reset the `EOC` before that line. For whatever reason, that did not happen

Thnaks @adamgreig for helping debug this :), and for a more detailed log of the steps we used to track it down, see the commit message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Fix PWM on `TIM1`
+- Fix ADC race condition causing incorrect reads at certain frequencies
 
 ## [v0.5.3] - 2020-01-20
 
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -349,13 +349,34 @@ macro_rules! adc_hal {
                     self.rb.sqr1.modify(|_, w| w.l().bits((len-1) as u8));
                 }
 
+                /**
+                  Performs an ADC conversion
+
+                  NOTE: Conversions can be started by writing a 1 to the ADON
+                  bit in the `CR2` while it is already 1, and no other bits
+                  are being written in the same operation. This means that
+                  the EOC bit *might* be set already when entering this function
+                  which can cause a read of stale values
+
+                  The check for `cr2.swstart.bit_is_set` *should* fix it, but
+                  does not. Therefore, ensure you do not do any no-op modifications
+                  to `cr2` just before calling this function
+                */
                 fn convert(&mut self, chan: u8) -> u16 {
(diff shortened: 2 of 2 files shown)


TheZoq2:
I forgot to mention that we think this commit fixes the symptoms of this issue, but it is still possible for it to occur.

There is no way to tell if the ADC is currently doing a conversion, so if a conversion is accidentally started *right before* calling `convert`, it might not be done when the `EOC` is cleared on the first line.

PR #204, opened 2020-04-23
rustfmt

torkeldanielsson:
I suggest using cargo fmt (rustfmt) to format the code.
Or is there some reason this is not done in this crate?

(This PR only does cargo fmt.)

### DIFF
--- a/examples/adc-dma-circ.rs
+++ b/examples/adc-dma-circ.rs
@@ -7,13 +7,8 @@ use panic_halt as _;
 
 use cortex_m::{asm, singleton};
 
-use stm32f1xx_hal::{
-    prelude::*,
-    pac,
-    adc,
-    dma::Half,
-};
 use cortex_m_rt::entry;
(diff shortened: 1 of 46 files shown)


therealprof:
@torkeldanielsson Well, rustfmt and it's capabilities changed quite a bit over time (e.g. the abilitiy to (partially) format within macro blocks). Also some people are not too excited about the default rustfmt formatting rules for PAC uses.

For me this is a clear improvement and I'd be in favour of that but I'll let other people with a stronger opinion chime in and reach a verdict on this PR.

TheZoq2:
Sounds like a good change. Most fmt changes look reasonable and having a consistent style is nice ~~(even though I hate the tie fighter `else`)~~ As @TeXitoi said, it's probably best if CI checks if the formatting is done correctly

therealprof:
For me doing that seems a good follow up step. Since we're all in agreement, let's get this merged.

Issue #109, opened 2019-09-06
Support for the STM32F107?


Rua:
I will be getting this board and I'd like to have support for it. Given its current lack, I had a look at what would be needed and if I could write basic support myself. The clock circuitry (RCC) is the biggest difference, as there's two dividers and two PLLs instead of just the single PLL like on the lower-end MCUs. This makes the `freeze` function incredibly complex. The number of possible combinations is huge, and finding the optimal one for the given input clock and desired sysclk becomes much more calculation-intensive.

I wonder if others have looked into this, and whether there is a workable solution?

therealprof:
I was planning on adding support a while ago due to this being the simplest STM32 MCU with Ethernet support but I never come around to it and put the project on ice. I haven't looked into it too much but I also feel this is going to be tricky.

nbraud:
Commenting here because I have some Olimex boards with that SoC (and was planning to send @Rua one).  I'll probably help her sort out how to compute clock parameters (fastest way might be to solve a diophantine system).

However, this situation suggests a couple of changes to the HAL crate:
- clarify whether `freeze()` is expected to exactly fulfil the desired clock rates: it's not always possible, and perhaps there should be two functions (one which returns `Result<Clocks, ClockError>` and is exact, and one which cannot fail but may be approximate?) ;
- separate the computation of clock/pll/... parameters into a `const fn`: in the most common case, the desired clock frequencies are known at compile time and so the parameters can be precomputed.

TheZoq2:
> clarify whether freeze() is expected to exactly fulfil the desired clock rates: it's not always possible, and perhaps there should be two functions (one which returns Result<Clocks, ClockError> and is exact, and one which cannot fail but may be approximate?) ;

The current implementation approximates it which I'm not a huge fan of. Though I'd almost argue that a panic is more appropriate than a Result.

> separate the computation of clock/pll/... parameters into a const fn: in the most common case, the desired clock frequencies are known at compile time and so the parameters can be precomputed.

That's not a bad idea. Is constfn stable enough that we can do everything we need? Last I checked they are pretty limited

TheZoq2:
> I'm not sure what makes more sense if there is no exact configuration. Result gives the user the freedom to decide what they want to do, which panic doesn't. Approximating seems ok, but perhaps that should be a choice the user makes rather than always doing it.

[...]

PR #205, opened 2020-04-24
Support for connectivity line devices (stm32f105 and stm32f107)

timokroeger:
This PR adds support for the `stm32f105xx` and `stm32f107xx` devices which are also knows as *connectivity line*.
They have a peripheral configuration similar to the stm32f103 high-density devices but feature USB OTG FS peripheral and the stm32f107xx devices additionally have an ethernet controller.

Both share a common SVD file with the name stm32f107. I introduced the `connectivity` cargo feature to always cover both devices. The `connectivity` feature replaces all previously abandoned `stm32f107` and `stm32f105` features in configuration attributes.

I could blink a LED on a STM32F105 board with a crystal oscillator. No other tests were run on real hardware.

Please have a look and let me know if I missed something and what can be improved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,8 @@ jobs:
           - stm32f100
           - stm32f101
           - stm32f103
+          - stm32f105
+          - stm32f107
         rust:
           - stable
         include:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Support for connectivity line devices: `stm32f105xx` and `stm32f107xx`
 - Consistently use PAC as `pac` and mark `device` and `stm32` informally as deprecated
 - Replace default blocking spi Write implementation with an optimized one
 - Use `Deref` for SPI generic implementations instead of macros
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,6 +86,8 @@ rt = ["stm32f1/rt"]
 stm32f100 = ["stm32f1/stm32f100", "device-selected"]
 stm32f101 = ["stm32f1/stm32f101", "device-selected"]
 stm32f103 = ["stm32f1/stm32f103", "device-selected"]
+stm32f105 = ["stm32f1/stm32f107", "device-selected", "connectivity"]
+stm32f107 = ["stm32f1/stm32f107", "device-selected", "connectivity"]
 
 # Devices with 64 or 128 Kb ROM
 medium = []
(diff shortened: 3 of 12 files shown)


burrbull:
LGTM

timokroeger:
I found one last issue when building for the existing targets.
For the connectivity line devices the cfgr `pllmul().bits()` method is unsafe.
This leads to an ``unnecessary `unsafe` block`` compiler warning when compiling for stm32f103 where the same method is safe.

Is there a better way to fix this than adding yet another `#[cfg()]` attribute?

timokroeger:
`#[allow(unused_unsafe)]` fixed the last remaining warning.
This PR is now ready to merge.

TheZoq2:
Thanks for the PR, looks good to me.

I wonder if the recent version bump to stm32-rs (#206) will have fixed that unused unsafe thing. If not, I'd say we should put a comment explaining why it is there.

burrbull:
This is because f105/107 have less than 16 levels for `pllmul`:
https://docs.rs/stm32f1/0.11.0/stm32f1/stm32f107/rcc/cfgr/struct.PLLMUL_W.html

TheZoq2:
Ah alright, I guess a comment is appropriate then

Rua:
Great work! I'm very curious how you managed to work out the settings for the PLLs and dividers for the user-requested Hz, since there's so many degrees of freedom and possible combinations on the 107.

timokroeger:
Rua you rised a very good point!
I assumed the stm32f103 clock configuration to be portable. I have to investigate if this is correct and need to check if the code can be reused.
Anyway with the recent commits to master this PR needs a rebase before it can land.

Rua:
Clock config is very different on the 107, see #109.

therealprof:
@Rua Yes it is different and even ST themselves got it wrong for quite a while... https://github.com/zephyrproject-rtos/zephyr/pull/9300

timokroeger:
stm32f105/107 only allow PLLMUL values 4..=9
All other stm32f10x devices support 2..=16
With the patch I just put a restriction on that range. That means the number of possible configurations dropped significantly especially if the internal oscillator is used. The maximum possible sysclk in that case is 9*8MHz/2=36MHz.
Yet it is sufficient for my use case and probably enough for to be useful in general.
ATM I do not have the time and knowledge to implement a more sophisticated logic to choose clock frequencies.

therealprof:
The trickier part is getting the frequencies right if you want to use Ethernet and USB. Anyway this needs a rebase to get rid of the merge conflicts. Probably easiest if you run `cargo fmt` locally before rebasing.

timokroeger:
Squashed and rebased

PR #207, opened 2020-04-26
Attempt to use GH actions for rustfmt check

therealprof:


therealprof:
As agreed on in #204 I'm going to set up automatic rustfmt checks.

PR #209, opened 2020-04-26
Fix clippy lints

therealprof:
Signed-off-by: Daniel Egger <daniel@eggers-club.de>

### DIFF
--- a/src/rtc.rs
+++ b/src/rtc.rs
@@ -18,7 +18,6 @@ use crate::backup_domain::BackupDomain;
 use crate::time::Hertz;
 
 use core::convert::Infallible;
-use nb;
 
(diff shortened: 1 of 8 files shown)


TheZoq2:
LGTM

src/rtc.rs
```diff
@@ -180,7 +179,7 @@ impl Rtc {
       ```
     */
     pub fn wait_alarm(&mut self) -> nb::Result<(), Infallible> {
-        if self.regs.crl.read().alrf().bit() == true {
```
burrbull, line 183: It's better use `bit_is_set/bit_is_clear()` here`

therealprof: I agree, didn't think of it; just wanted to do a quick cleanup to shut up clippy. 😅

PR #213, opened 2020-05-03
Rework examples to apply to more targets and test in CI

therealprof:
This also improves the required-features selection for the examples.

Signed-off-by: Daniel Egger <daniel@eggers-club.de>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,7 @@
-on: [push, pull_request]
+on:
+  push:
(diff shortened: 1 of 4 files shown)


burrbull:
`bors` does not manage this repository.

We should fix this.

therealprof:
> bors does not manage this repository.
> We should fix this.

We could, however there's very little benefit for small teams and even worse it has it's own set of funny issues.

TheZoq2:
Sorry for the delay! Looks like some good changes :)

PR #217, opened 2020-05-17
Add blinky_rtcalarm_irq example

khrs:
It is not obvious how to use RTCALARM interrupt. It took me a near an hour to find it in reference manual.

### DIFF
--- a/examples/blinky_rtcalarm_irq.rs
+++ b/examples/blinky_rtcalarm_irq.rs
@@ -1,0 +1,114 @@
+//! Blinks an LED
+//!
+//! This assumes that a LED is connected to pc13 as is the case on the blue pill board.
+//!
+//! Note: Without additional hardware, PC13 should not be used to drive a LED, see
+//! section 5.1.2 of the reference manual for an explanation.
+//! This is not an issue on the blue pill.
+
+#![no_std]
+#![no_main]
+
+use panic_halt as _;
+
+use stm32f1xx_hal as hal;
+
+use crate::hal::{
+    gpio::*,
+    pac::{interrupt, Interrupt, Peripherals, EXTI},
+    prelude::*,
+    rtc::*,
+};
+
+use core::cell::RefCell;
+use cortex_m::{asm::wfi, interrupt::Mutex, peripheral::Peripherals as c_m_Peripherals};
+use cortex_m_rt::entry;
+use embedded_hal::digital::v2::OutputPin;
+
+// A type definition for the GPIO pin to be used for our LED
+type LEDPIN = gpioc::PC13<Output<PushPull>>;
+
+// Make LED pin globally available
+static G_LED: Mutex<RefCell<Option<LEDPIN>>> = Mutex::new(RefCell::new(None));
+// Make RTC globally available
+static G_RTC: Mutex<RefCell<Option<Rtc>>> = Mutex::new(RefCell::new(None));
+// Make EXTI registers globally available
+static G_EXTI: Mutex<RefCell<Option<EXTI>>> = Mutex::new(RefCell::new(None));
+
+// Toggle LED every 3 seconds
+const TOGGLE_INTERVAL_SECONDS: u32 = 3;
+
+#[interrupt]
+fn RTCALARM() {
+    static mut LED: Option<LEDPIN> = None;
+    static mut RTC: Option<Rtc> = None;
+    static mut EXTI: Option<EXTI> = None;
+
+    let led = LED.get_or_insert_with(|| {
+        cortex_m::interrupt::free(|cs| {
+            // Move LED pin here, leaving a None in its place
+            G_LED.borrow(cs).replace(None).unwrap()
+        })
+    });
+    let rtc = RTC.get_or_insert_with(|| {
+        cortex_m::interrupt::free(|cs| G_RTC.borrow(cs).replace(None).unwrap())
+    });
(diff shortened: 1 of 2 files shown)


examples/blinky_rtcalarm_irq.rs
```diff
@@ -0,0 +1,114 @@
+//! Blinks an LED
+//!
+//! This assumes that a LED is connected to pc13 as is the case on the blue pill board.
+//!
+//! Note: Without additional hardware, PC13 should not be used to drive a LED, see
+//! section 5.1.2 of the reference manual for an explanation.
+//! This is not an issue on the blue pill.
```
TheZoq2, line 7: Would you mind adding some comment about this using RTC and IRQ here?

examples/blinky_rtcalarm_irq.rs
```diff
@@ -0,0 +1,114 @@
...
+
+use panic_halt as _;
+
+use stm32f1xx_hal as hal;
+
+use crate::hal::{
+    gpio::*,
+    pac::{interrupt, Interrupt, Peripherals, EXTI},
+    prelude::*,
+    rtc::*,
```
TheZoq2, line 20: I'm not a huge fan of glob imports (apart from `prelude`)

examples/blinky_rtcalarm_irq.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    exti.pr.write(|w| w.pr17().set_bit());
+    rtc.set_alarm(rtc.current_time() + TOGGLE_INTERVAL_SECONDS);
+
+    let _ = led.toggle();
+}
+
+#[entry]
+fn main() -> ! {
+    if let (Some(dp), Some(cp)) = (Peripherals::take(), c_m_Peripherals::take()) {
```
TheZoq2, line 68: I prefer unwrapping `dp` and `cp` as we never expect that to fail, and it would be a fatal error if it does

examples/blinky_rtcalarm_irq.rs
```diff
@@ -0,0 +1,114 @@
...
+            rtc.listen_alarm();
+
+            *G_RTC.borrow(cs).borrow_mut() = Some(rtc);
+
+            // Enable RTCALARM IRQ, set prio 1 and clear any pending IRQs
+            let mut nvic = cp.NVIC;
+            // Calling 'set_priority()' and 'unmask()' requires 'unsafe {}'
+            // - https://docs.rs/stm32f1xx-hal/0.5.3/stm32f1xx_hal/stm32/struct.NVIC.html#method.set_priority
+            unsafe {
+                nvic.set_priority(Interrupt::RTCALARM, 1);
```
TheZoq2, line 103: I think the default priority would be fine, right?

examples/blinky_rtcalarm_irq.rs
```diff
@@ -0,0 +1,114 @@
...
+                cortex_m::peripheral::NVIC::unmask(Interrupt::RTCALARM);
+            }
+            // Mark as pending
+            cortex_m::peripheral::NVIC::pend(Interrupt::RTCALARM);
```
TheZoq2, line 107: If we start the RTC with the right time, it shouldn't be nessecary to manually pend the interrupt

examples/blinky_rtcalarm_irq.rs
```diff
@@ -0,0 +1,114 @@
...
+    exti.pr.write(|w| w.pr17().set_bit());
+    rtc.set_alarm(rtc.current_time() + TOGGLE_INTERVAL_SECONDS);
+
+    let _ = led.toggle();
+}
+
+#[entry]
+fn main() -> ! {
+    if let (Some(dp), Some(cp)) = (Peripherals::take(), c_m_Peripherals::take()) {
+        cortex_m::interrupt::free(move |cs| {
```
TheZoq2, line 69: We don't really need the interrupt free until we start turning on interrupts, right?

examples/blinky_rtcalarm_irq.rs
```diff
@@ -0,0 +1,114 @@
...
+            let mut gpioc = dp.GPIOC.split(&mut rcc.apb2);
+            let mut led = gpioc.pc13.into_push_pull_output(&mut gpioc.crh);
+            let _ = led.set_high(); // Turn off
+
+            *G_LED.borrow(cs).borrow_mut() = Some(led);
+
+            // Set up the EXTI (see notes in section 18.4.2 of reference manual)
+            let exti = dp.EXTI;
+            exti.ftsr.write(|w| w.tr17().set_bit());
+            exti.imr.write(|w| w.mr17().set_bit());
```
TheZoq2, line 83: I *think*, based on reading the mentioned section of the RM, that this isn't required since we are listening for the RTCALARM interrupt, rather than the exti interrupt

khrs: This is why i spend near hour ;)

rm0008-p489

khrs: RM008 page 489

TheZoq2: Right, I read that yesterday but think I missinterpreted it. Perhaps we should add a convenience that does this at some point :thinking:

TheZoq2:
Looks great now, thanks. Unfortunately, the build fails on the f101 which I presume is  because the f101 doesn't have the RTC. I guess we have to add some config to not build the example if the 101 feature is selected

PR #218, opened 2020-05-19
Add missing watchdog prescalers.

khrs:
According to RM0008 section 25.4.2 we should support prescaler values up to 256.


TheZoq2:
Good catch, thanks!

PR #221, opened 2020-05-24
Cleanup blinky_timer_irq example.

### LINKED ISSUE #220: blinky_timer_irq could use some improvements
As mentioned in #217, there are some issues with the [blinky_timer_irq] example:

- `if let` on peripherals instead of `unwrap`
- overly large interrupt free section

[blinky_timer_irq]: https://github.com/stm32-rs/stm32f1xx-hal/blob/master/examples/blinky_timer_irq.rs

khrs:
Fixes #220

### DIFF
--- a/examples/blinky_timer_irq.rs
+++ b/examples/blinky_timer_irq.rs
@@ -1,9 +1,11 @@
-// blinky timer using interrupts on TIM2
-//
-// This demo based off of the following demo:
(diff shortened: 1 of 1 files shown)


TheZoq2:
Oh nice, I half expected that issue to sit around for a long time. Thanks for fixing it!

PR #229, opened 2020-06-20
Improve blocking i2c ergonomics

TheZoq2:
Lumpio- in the community discord rightfully pointed out that our current blocking i2c struct takes a lot of unnamed parameters for initialisation. This replaces those unnamed parameter with a default-constructible struct, which allows either the builder pattern, or `Struct{fields, .. default()}` initialisation to be used.

I also renamed `retries` to `attempts` because I think retries is ambiguous. I interpret it as "how many extra tries should we make", but the implementation is "how many attempts should we make.

Finally, I moved the mapr reference requried for i2c1 to the end of the parameter list to the end, next to apb. This makes a lot more sense to me since most of our APIs takes those parameters last, and it's more compatible with i2c2

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -100,7 +100,7 @@ pub struct I2c<I2C, PINS> {
 pub struct BlockingI2c<I2C, PINS> {
     nb: I2c<I2C, PINS>,
     start_timeout: u32,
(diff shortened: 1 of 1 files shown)


TheZoq2:
Closing in favour of #231

PR #231, opened 2020-06-20
Use statically typed time consistently

### LINKED ISSUE #230: Statically typed time
We have some `time_us` variables laying around, for example in the blocking i2c implementation

I think it might be better to use the MicroSecond/MilliSecond structs defined in `time.rs` instead, for reducing the risk of errors, and making clear what the unit is at the call site

TheZoq2:
Closes #230 

Replaces the uses of `_ms` and `_us` postfix variables with `Microsecond` and `Millisecond` types. I also renamed the types to use a lowercase S which I believe is more correct. It's the way wikipedia spells it at least :) https://en.wikipedia.org/wiki/Microsecond

This is built on top of #229 as that was where a lot of the `us` occurrences were.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -100,7 +100,7 @@ pub struct I2c<I2C, PINS> {
 pub struct BlockingI2c<I2C, PINS> {
     nb: I2c<I2C, PINS>,
     start_timeout: u32,
-    start_retries: u8,
+    start_attempts: u8,
     addr_timeout: u32,
     data_timeout: u32,
 }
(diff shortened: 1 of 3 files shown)


TheZoq2:
Oh, and I want to do another patch release with some doc improvements and some non-breaking PRs, so we should probably hold off on merging this until that is done.

src/i2c.rs
```diff
@@ -1,16 +1,17 @@
-//! Inter-Integrated Circuit (I2C) bus
```
therealprof, line 1: What's the reason for the comment style change? C++ comments are not so typical in the Rust world.

TheZoq2: Oops, I think I started adding a larger doc chunk, but then changed my mind about what I wanted to do. I'll revert it until I get around to adding i2c docs

therealprof: Even then it should be possible to use Rust style comments. ;)

TheZoq2: Sure, but I kind of dislike that for large blocks, especially inserting new lines

TheZoq2:
Of course, I can't just break the tradition of forgetting to include changelogs with changes :)

burrbull:
Can we use this?
https://docs.rs/embedded-time/0.5.0/embedded_time/index.html

TheZoq2:
Yes, that's probably a better idea than using our own structs. I'll port our code over to that

TheZoq2:
Looks like using unsigned base types with `embedded_time` is a bit tricky https://github.com/FluenTech/embedded-time/issues/19

PR #232, opened 2020-06-20
Remove generic RCC buses from public API

TheZoq2:
I never liked the RccBus generic stuff, as it made it impossible to tell which register to pass to the constructors of functions, `&mut <USART1 as RccBus>::Bus` isn't very helpful.

This PR goes back to explicit APB1 and APB2 in the constructors, but keeps the generic structs inside functions to still provide most of the benefit while improving doc readbility.

While this is a public API change, I don't believe it is breaking, as it still requires the exact same types to be passed to functions.

cc: @burrbull


burrbull:
Ok

PR #233, opened 2020-06-21
Improve SPI, RCC and AFIO documentation

TheZoq2:
The lack of documentation for how to get a hold of the `Rcc` and `Afio` structs has bothered me for a while. This fixes that, along with adding some top level documentation to the SPI module

### DIFF
--- a/src/afio.rs
+++ b/src/afio.rs
@@ -33,6 +33,15 @@ impl AfioExt for AFIO {
     }
 }
 
+/// HAL wrapper around the AFIO registers
+///
+/// Aquired by calling [constrain](trait.AfioExt.html#constrain) on the [AFIO
+/// registers](../pac/struct.AFIO.html)
+///
+/// ```rust
+/// let p = pac::Peripherals::take().unwrap();
+/// let mut rcc = p.RCC.constrain();
+/// let mut afio = p.AFIO.constrain(&mut rcc.apb2);
 pub struct Parts {
     pub evcr: EVCR,
     pub mapr: MAPR,
@@ -53,6 +62,16 @@ impl EVCR {
     }
 }
 
+/// AF remap and debug I/O configuration register (MAPR)
+///
+/// Aquired through the [Parts](struct.Parts.html) struct.
+///
+/// ```rust
+/// let dp = pac::Peripherals::take().unwrap();
+/// let mut rcc = dp.RCC.constrain();
+/// let mut afio = dp.AFIO.constrain(&mut rcc.apb2);
+/// function_using_mapr(&mut afio.mapr);
+/// ```
 pub struct MAPR {
     _0: (),
     jtag_enabled: bool,
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -48,6 +48,14 @@ pub struct Rcc {
 }
 
 /// AMBA High-performance Bus (AHB) registers
+///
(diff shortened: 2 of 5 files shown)


src/afio.rs
```diff
@@ -33,6 +33,15 @@ impl AfioExt for AFIO {
     }
 }
 
+/// HAL wrapper around the AFIO registers
+///
+/// Aquired by calling [constrain](trait.AfioExt.html#constrain) on the [AFIO
```
therealprof, line 38:
```suggestion
/// Acquired by calling [constrain](trait.AfioExt.html#constrain) on the [AFIO
```

src/afio.rs
```diff
@@ -33,6 +33,15 @@ impl AfioExt for AFIO {
...
+/// registers](../pac/struct.AFIO.html)
+///
+/// ```rust
```
therealprof, line 41: `rust` is superfluous here as it is automatically assumed, but doc tests for embedded should probably at least feature the `no_run` attribute (if they **can** be compiled but will not run on the host) or even `ignore` if the example cannot compile due to ficticious code.

src/rcc.rs
```diff
@@ -48,6 +48,14 @@ pub struct Rcc {
 }
 
 /// AMBA High-performance Bus (AHB) registers
+///
+/// Aquired through the `Rcc` registers:
```
therealprof, line 52:
```suggestion
/// Acquired through the `Rcc` registers:
```

src/rcc.rs
```diff
@@ -62,6 +70,14 @@ impl AHB {
 }
 
 /// Advanced Peripheral Bus 1 (APB1) registers
+///
+/// Aquired through the `Rcc` registers:
```
therealprof, line 74:
```suggestion
/// Acquired through the `Rcc` registers:
```

src/rcc.rs
```diff
@@ -86,6 +102,14 @@ impl APB1 {
 }
 
 /// Advanced Peripheral Bus 2 (APB2) registers
+///
+/// Aquired through the `Rcc` registers:
```
therealprof, line 106:
```suggestion
/// Acquired through the `Rcc` registers:
```

src/rcc.rs
```diff
@@ -169,6 +203,19 @@ impl CFGR {
         self
     }
 
+    /// Applies the clock configuration and returns a `Clocks` struct that signifies that the
+    /// clocks are frozen, and contains the frequencies used. After this function is called,
```
therealprof, line 207:
```suggestion
    /// clocks are frozen and contains the frequencies used. After this function is called,
```

src/afio.rs
TheZoq2: Oh right, I almost forgot doctests are a thing. It would be nice to them where possible, though I feel like it might require a fair amount of extra code in a lot of docstrings.

TheZoq2: Since it seems to be kind of difficult to run doctests right now, I'd say we should merge this, release a new version to get updated docs on docs.rs, and then sort out doctests later.

Any objections?

therealprof: There's a conflict though.

TheZoq2: Oh right, there are always changelog problems :) I'll merge #183 then sort these out and merge this

PR #235, opened 2020-06-22
Fix period retrival for timers

fooker:
Fix `get_period` for timers


TheZoq2:
Oh, that's a sneaky bug, good catch! :)

Would you mind also adding a changelog entry to `CHANGELOG.md`?

fooker:
> Oh, that's a sneaky bug, good catch! :)
> 
> Would you mind also adding a changelog entry to `CHANGELOG.md`?

Done.

PR #239, opened 2020-06-27
Improve DMA ergonomics when working with generics

TheZoq2:
I found our current DMA API to be quite difficult to work with in external crates because of the `A, B` thingy. This replaces the buffer types with the ones used in the embeddonomicon https://docs.rust-embedded.org/embedonomicon/dma.html

As  far as I can tell, this is safe, but I may very well be missing something.

Because of the way that `Transfer` is implemented, I found no way to specify a generic TxDma which has a `wait` method. To fix that, I added the `Transferable` trait.

Example usage of all of this can be found at https://gitlab.com/TheZoq2/ws2812-spi-dma/-/blob/master/src/lib.rs

DNM because the same treatment has to be given to RxDma, but I just want to make sure I don't do anything stupid before going through with that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Breaking changes
+
+- Replace DMA buffer types with `Deref` ans `Unpin`
+
 ## [v0.6.1] - 2020-06-25
 
 ### Added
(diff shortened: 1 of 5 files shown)


thalesfragoso:
You might want to wait on the resolution of https://github.com/rust-embedded/embedonomicon/issues/64

TheZoq2:
Oh, hadn't seen that. I'll have a look at that tomorrow, thanks for the heads up!

TheZoq2:
I added StableDeref now, hopefully I did that correctly :sweat_smile:

TheZoq2:
Closing this in favor of #274

PR #241, opened 2020-06-30
Change arm-none-eabi-gdb to gdb-multiarch

rudihorn:
On Ubuntu from 18.04 it seems that `arm-none-eabi-gdb` is now replaced with `gdb-multiarch` and then `gdb` is used to perform debugging. https://rust-embedded.github.io/book/intro/install/linux.html seems to indicate this might not apply to all distributions though. A `gdb-multiarch` package does exist in the AUR (https://aur.archlinux.org/packages/gdb-multiarch/), and I'm unable to access the fedora website to check.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,5 +1,5 @@
 [target.thumbv7m-none-eabi]
-runner = 'arm-none-eabi-gdb'
+runner = 'gdb'
 rustflags = [
   "-C", "link-arg=-Tlink.x",
 ]
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,5 +1,5 @@
 [target.thumbv7m-none-eabi]
-runner = 'gdb'
+runner = 'gdb-multiarch'
 rustflags = [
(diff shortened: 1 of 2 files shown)


TheZoq2:
Since this seems to depend on the distribution, would it be better to have something like "use arm-none-eabi-gdb or gdb-multiarch depending on what is easiest to install"?

rudihorn:
> 
> 
> Since this seems to depend on the distribution, would it be better to have something like "use arm-none-eabi-gdb or gdb-multiarch depending on what is easiest to install"?

For `README.md` this would make sense, and be documented appropriately.

For `.cargo/config` I assume we have to have one value for the runner, and I think given the more widespread usage of Debian/Ubuntu based distributions I would suggest it being gdb-multiarch.

TheZoq2:
Yea, that sounds very reasonable! Would you mind updating the PR with that change?

TheZoq2:
Perfect, thanks a lot!

Issue #212, opened 2020-05-02
Support for ADC multichannel continuous conversion?


yjh0502:
Hi, I'm trying to run a ADC multichannel (scan mode) continuous conversion. I dig into the code to check current status:

`Continuous` conversion: set `cont()` bit on `start()`
https://github.com/stm32-rs/stm32f1xx-hal/blob/89600ec0708cc9cb66c72c1b2c4b6b3a4827c7cc/src/adc.rs#L558-L568

`Scan` conversion: not setting `cont()` bit on `start()` clear `cont()` explicitly on initialization. It is definitely not intended for continuous conversion.
https://github.com/stm32-rs/stm32f1xx-hal/blob/89600ec0708cc9cb66c72c1b2c4b6b3a4827c7cc/src/adc.rs#L608-L617

I think there are several ways to support continuous scan conversion mode:

 - I could add quick and dirty fix by introducing `ContinuousScan` struct (or `struct Scan { continuous: bool }` if it's okay to break API), but I'm not sure if it's a right way.
 - I could think of generalized struct/trait which covers single/multichannel one-shot/continuous conversion scenario. Single-channel conversion could be treated as a special case of multichannel conversion anyway.

TheZoq2:
I'm not quite sure what you're asking, as far as I remember, the DMA should do continuous conversion, or does it just do multiple conversions on the same channel?

I think breaking the API is perfectly fine, we do that quite often.

yjh0502:
https://www.st.com/resource/en/application_note/cd00258017-stm32s-adc-modes-and-their-applications-stmicroelectronics.pdf

![image](https://user-images.githubusercontent.com/554344/80869171-c3439700-8cd9-11ea-8b7c-c01503c90ac3.png)

I mean `Scan` mode on `stm32f1xx-hal` API with continuous conversion.

On current API, we should choose between
 - single-channel single-shot conversion (`OneShot::read` API)
 - single-channel continuous conversion with DMA (`Adc::with_dma`, which returns `Continuous` mode) or,
 - multi-channel single-shot conversion with DMA (`Adc::with_scan_dma`, which returns `Scan` mode, scanning multiple channels by order, but only once)

[...]

PR #247, opened 2020-07-15
Added support for continuous and discontinuous scan mode

katyo:
Adds support for continuous scan mode (solves issue #212).
Also adds support for discontinuous conversion.

Different approach than #214. This PR does not break existing API.


TheZoq2:
Thanks for the PR and sorry to keep you waiting for so long for the review!

The changes look good to me, would you mind adding a changelog entry?

Also, cc @yjh0502 since you were the author of the other linked PR, do you have any opinion on this? Does it suit your use case?

PR #249, opened 2020-07-20
More documentation links

TheZoq2:
Like my previous documentation improvement PR, this one adds a few more links where I think that it might be hard to find out where parameters come from, or how to aquire certain structs.

I'll probably do a few more commits as I find them, so don't merge quite yet


TheZoq2:
I didn't run into any more things like this, so this is ready to merge.

cc @therealprof

PR #251, opened 2020-07-26
Adapt to embedded-hal 1.0.0-alpha.1

eldruin:
Blocked on https://github.com/rust-embedded/embedded-hal/issues/177.
Currently it does not build because the `mfrc522` crate used in an example needs to be adapted as well. Although this driver seems to be a bit out of date.


TheZoq2:
Awesome, I gave it a quick overlook but should probably read it a bit more thoroughly before we merge. Does it make sense to port to the alpha version or should we wait for a full release? I'm inclined to do the latter, especially since I assume most drivers aren't up to date yet.

Sidenote: I believe this has been discussed in other places, but is it really nessecary to have `try` everywhere. It seems redundant since we see the potential issue at the type level, and there is no infallible things. Seems to just be extra clutter to me now...

eldruin:
Good question. I think it is a bit of a chicken-egg problem with HALs and drivers but for now I would wait for the final release since discussions at `embedded-hal` have picked up pace lately.
If you want to make a release, I would recommend making it also an `-alpha.1` so that cargo does not upgrade it automatically.
Since there are so many breaking changes I thought it would help the community to prepare at least most of the changes already, so that it takes shorter to move the ecosystem to `e-h` `1.0.0`.

There was a meeting where it was decided to make all methods fallible. See [here](https://hackmd.io/ck-xRXtMTmKYXdK5bEh82A?view#Final-Summary) for the reasoning.

therealprof:
> Sidenote: I believe this has been discussed in other places, but is it really nessecary to have `try` everywhere. It seems redundant since we see the potential issue at the type level, and there is no infallible things. Seems to just be extra clutter to me now...

Well, Rust also uses `try` a lot and similarly to Rust where a lot of things are available in fallible and infallible versions some people would love to see infallible traits to make it back, too. For me labelling those methods `try` is really about consistency and allowing to have infallible things without the `digital::v1` vs. `digital::v2` kerfuffle.

TheZoq2:
Oh right, the point of leaving the possibility for infallible traits is a good one :)

eldruin:
Closing in favor of #296

PR #257, opened 2020-08-16
Flash Read/Write Support

irwineffect:
This PR is a followup on #141.

### DIFF
--- a/src/flash.rs
+++ b/src/flash.rs
@@ -2,6 +2,294 @@
 
 use crate::pac::{flash, FLASH};
 
+pub const FLASH_START: u32 = 0x0800_0000;
+pub const FLASH_END: u32 = 0x080F_FFFF;
+
+const _RDPRT_KEY: u16 = 0x00A5;
+const KEY1: u32 = 0x45670123;
+const KEY2: u32 = 0xCDEF89AB;
+
+pub const SZ_1K: u16 = 1024;
+
+pub type Result<T> = core::result::Result<T, Error>;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
+pub enum Error {
+    AddressLargerThanFlash,
+    AddressMisaligned,
+    LengthNotMultiple2,
+    LengthTooLong,
+    EraseError,
+    ProgrammingError,
+    WriteError,
+    VerifyError,
+    UnlockError,
+    LockError,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
+pub enum SectorSize {
+    Sz1K = 1,
+    Sz2K = 2,
+    Sz4K = 4,
+}
+impl SectorSize {
+    const fn kbytes(self) -> u16 {
+        SZ_1K * self as u16
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd)]
+pub enum FlashSize {
+    Sz16K = 16,
+    Sz32K = 32,
+    Sz64K = 64,
+    Sz128K = 128,
+    Sz256K = 256,
+    Sz384K = 384,
+    Sz512K = 512,
+    Sz768K = 768,
+    Sz1M = 1024,
+}
+impl FlashSize {
+    const fn kbytes(self) -> u32 {
+        SZ_1K as u32 * self as u32
+    }
+}
+
+pub struct FlashWriter<'a> {
+    flash: &'a mut Parts,
+    sector_sz: SectorSize,
+    flash_sz: FlashSize,
+    verify: bool,
+}
+impl<'a> FlashWriter<'a> {
+    fn unlock(&mut self) -> Result<()> {
+        // Wait for any ongoing operations
+        while self.flash.sr.sr().read().bsy().bit_is_set() {}
+
+        // NOTE(unsafe) write Keys to the key register. This is safe because the
+        // only side effect of these writes is to unlock the flash control
+        // register, which is the intent of this function. Do not rearrange the
+        // order of these writes or the control register will be permanently
+        // locked out until reset.
+        unsafe {
+            self.flash.keyr.keyr().write(|w| w.key().bits(KEY1));
+        }
+        unsafe {
+            self.flash.keyr.keyr().write(|w| w.key().bits(KEY2));
+        }
+
+        // Verify success
+        match self.flash.cr.cr().read().lock().bit_is_clear() {
+            true => Ok(()),
+            false => Err(Error::UnlockError),
+        }
+    }
+
+    fn lock(&mut self) -> Result<()> {
+        //Wait for ongoing flash operations
+        while self.flash.sr.sr().read().bsy().bit_is_set() {}
+
+        // Set lock bit
+        self.flash.cr.cr().modify(|_, w| w.lock().set_bit());
+
+        // Verify success
+        match self.flash.cr.cr().read().lock().bit_is_set() {
+            true => Ok(()),
+            false => Err(Error::LockError),
+        }
+    }
+
+    fn valid_address(&self, offset: u32) -> Result<()> {
+        if FLASH_START + offset > FLASH_END {
+            Err(Error::AddressLargerThanFlash)
+        } else if offset & 0x1 != 0 {
+            Err(Error::AddressMisaligned)
+        } else {
+            Ok(())
+        }
+    }
+
+    fn valid_length(&self, offset: u32, length: usize) -> Result<()> {
+        if offset + length as u32 > self.flash_sz.kbytes() as u32 {
+            Err(Error::LengthTooLong)
+        } else if length & 0x1 != 0 {
+            Err(Error::LengthNotMultiple2)
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Erase sector which contains start_offset
+    pub fn page_erase(&mut self, start_offset: u32) -> Result<()> {
+        self.valid_address(start_offset)?;
+
+        // Unlock Flash
+        self.unlock()?;
+
+        // Set Page Erase
+        self.flash.cr.cr().modify(|_, w| w.per().set_bit());
+
+        // Write address bits
+        // NOTE(unsafe) This sets the page address in the Address Register.
+        // The side-effect of this write is that the page will be erased when we
+        // set the STRT bit in the CR below. The address is validated by the
+        // call to self.valid_address() above.
+        unsafe {
+            self.flash
+                .ar
+                .ar()
+                .write(|w| w.far().bits(FLASH_START + start_offset));
+        }
+
+        // Start Operation
+        self.flash.cr.cr().modify(|_, w| w.strt().set_bit());
+
+        // Wait for operation to finish
+        while self.flash.sr.sr().read().bsy().bit_is_set() {}
+
+        // Check for errors
+        let sr = self.flash.sr.sr().read();
+
+        // Remove Page Erase Operation bit
+        self.flash.cr.cr().modify(|_, w| w.per().clear_bit());
+
+        // Re-lock flash
+        self.lock()?;
+
+        if sr.wrprterr().bit_is_set() {
+            self.flash.sr.sr().modify(|_, w| w.wrprterr().clear_bit());
+            Err(Error::EraseError)
+        } else {
+            if self.verify {
+                // By subtracting 1 from the sector size and masking with
+                // start_offset, we make 'start' point to the beginning of the
+                // page, and similarly 'end' at the end of the page. We do this
+                // because the entire page should have been erased, regardless
+                // of where in the page the given 'start_offset' was.
+                let start = start_offset & !(self.sector_sz.kbytes() as u32 - 1);
+                let end = start + self.sector_sz.kbytes() as u32 - 1;
+                for idx in start..end {
+                    let write_address = (FLASH_START + idx as u32) as *const u16;
+                    let verify: u16 = unsafe { core::ptr::read_volatile(write_address) };
+                    if verify != 0xFFFF {
+                        return Err(Error::VerifyError);
+                    }
+                }
+            }
+
+            Ok(())
+        }
+    }
+
+    /// Erase the Flash Sectors from FLASH_START + start_offset to length
+    pub fn erase(&mut self, start_offset: u32, length: usize) -> Result<()> {
+        self.valid_length(start_offset, length)?;
+
+        // Erase every sector touched by start_offset + length
+        for offset in
+            (start_offset..start_offset + length as u32).step_by(self.sector_sz.kbytes() as usize)
+        {
+            self.page_erase(offset)?;
+        }
+
+        // Report Success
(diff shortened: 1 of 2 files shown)


burrbull:
Please, use `git rebase` instead of `merge` for PR.

irwineffect:
Alright, I've rebased on master, it also looks like I need to run rustfmt over the code?

src/flash.rs
```diff
@@ -20,6 +303,37 @@ impl FlashExt for FLASH {
...
+    pub(crate) optkeyr: OPTKEYR,
+
+    /// Opaque SR register
+    pub(crate) sr: SR,
+
+    /// Opaque WRPR register
+    pub(crate) wrpr: WRPR,
+}
+impl Parts {
+    pub fn writer(&mut self, sector_sz: SectorSize, flash_sz: FlashSize) -> FlashWriter {
```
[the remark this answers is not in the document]
TheZoq2: Unless I'm missing something, `&mut self` only lasts until the function returns, but the `FlashWriter` lasts much longer than that.

A `release` method sounds like the best option to me. We could probably also expose the functionality you talk about through the flash writer if required, right?

src/flash.rs
```diff
@@ -2,6 +2,282 @@
...
+            while self.flash.sr.sr().read().bsy().bit_is_set() {}
+
+            // Set Page Programming to 0
+            self.flash.cr.cr().modify(|_, w| w.pg().clear_bit() );
+
+            // Check for errors
+            if self.flash.sr.sr().read().pgerr().bit_is_set() {
+                self.flash.sr.sr().modify(|_, w| w.pgerr().clear_bit() );
+
+                self.lock()?;
```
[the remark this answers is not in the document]
TheZoq2: That also sounds like a good option. And my personal opinion is that as long as there are no very high performance impacts, we should have as much safety in internal APIs as outside. I don't think either of the solutions will incur any runtime cost (in release mode :))

src/flash.rs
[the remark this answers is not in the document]
irwineffect: I just tested, it won't let us call writer() again. This is because the FlashWriter struct contains a mutable reference to Parts.

irwineffect:
I believe I've addressed all of the review findings, with the exception of handling the lock/unlock, still working on that!

src/flash.rs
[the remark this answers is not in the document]
irwineffect: We definitely could expose the extra functionality through the FlashWriter if we really needed to, but I think it would make sense to have it separate. In any case, I think we can rely of reference lifetimes to ensure safe/correct usage around the FlashWriter, Parts, and the writer() function.

src/flash.rs
```diff
@@ -2,6 +2,303 @@
...
+impl FlashSize {
+    const fn kbytes(self) -> u32 {
+        SZ_1K as u32 * self as u32
+    }
+}
+
+pub struct FlashReaderGuard<'a> {
+    _flash: &'a FlashWriter<'a>,
+    data: &'a [u8],
+}
```
thalesfragoso, line 63:
```suggestion
pub struct FlashReaderGuard<'b, 'a: 'b> {
    _flash: &'b FlashWriter<'a>,
    data: &'b [u8],
}
```

thalesfragoso:
Actually, I don't think we need the `Guard` at all, since

```rust
pub fn read(&self, offset: u32, length: usize) -> Result<&[u8]>
```
should be the same as

```rust
pub fn read(&'c self, offset: u32, length: usize) -> Result<&'c [u8]>
```

So it has to extend the borrow as long as the output lives if my understanding is correct, see this playground example:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ce7f1c92470429693d99969d87f985d5

irwineffect:
@thalesfragoso, yeah, I think you're right, we don't need to have the separate ReaderGuard, since the type system will prevent us from mutably borrowing the FlashWriter until the data returned by read() is out of scope, and we need to mutably borrow the FlashWriter before we could do any operations that would invalidate the data.

src/flash.rs
irwineffect: So, no longer relevant because we'll be removing the FlashReaderGuard entirely, but I'm curious what is the purpose of creating this second lifetime specifier? I looked up the syntax of `'a: 'b` means "`'a` must last at least as long as `'b`", but what does that give us?

irwineffect:
So, I've been playing around with different ways to deal with the locking/unlocking of the flash, and I've found things I really don't like about each solution, I'd appreciate some input.

I started out trying to wrap the `Parts` inside a `FlashLockGuard` struct that requires calling `unlock()` to get access the registers that require the flash hardware to be unlocked, and then calling `lock()` upon drop of the `FlashLockGuard`. The issue with this method is that with the current API, the `lock()` operation can fail, and I believe you don't want things that can fail inside a drop function, right?

The trouble with the `do_with_unlock()` method that @TheZoq2  had suggested is that if the `operation()` fails we will return without calling `lock()`. We could modify the example you provided to always `lock()` even if `operation()` fails, but if lock also fails, which error should we return, the lock failure or the operation failure? We could make a combo failure, but that feels pretty ugly.

I think my greater issue is the fact that the current `lock()`/`unlock()`API can fail. While unlocking could reasonably fail, I don't know under what circumstances the `lock()`operation would actually fail, I personally have not seen this occur, and I haven't noticed anything in the flash documentation that suggests this is possible. Can we not trust that writing to the lock bit in the control register will actually cause the bit to be set? Perhaps locking always succeeds, and we can ignore the possibility of failure? I'm inclined to continue punting the possible failure upwards to the user (or  panic?) until this library has been more thoroughly tested and we are more confident this can't fail.

Also, I don't know if there really is a strong need to unlock/lock the flash around each write/erase operation. This does minimize the window that the flash could get corrupted if the code were to go off the rails, but why not just unlock the flash when the `FlashWriter` is created, keep it unlocked the entire time the `FlashWriter` exists, and then provide a `lock()` method that consumes the `FlashWriter` that the user can call when they are done with the flash? The only issue I foresee with this method is that the user could forgot to call `lock()` to turn back on the hardware lock on the flash (although is isn't the worst thing in the world). To protect against this possibility, we could implement a custom `Drop` for the `FlashWriter`, but this still has the issue that the locking operation could fail (but perhaps we should just ignore it or panic?). Also, if the user structures the program such that the FlashWriter is alive for the entire lifetime of the program, we are in the same boat as before. I think an explicit lock is better for that reason.

I'd love to get @luctius's thoughts to know the intent behind the original design, but barring that, what do you guys think?

src/flash.rs
```diff
@@ -219,12 +208,14 @@ impl<'a> FlashWriter<'a> {
...
-            // NOTE(unsafe) read with no side effects safe-guarded by FlashReaderGuard.
-            data: unsafe { core::slice::from_raw_parts::<'static, u8>(address, length) },
-        })
+        Ok(
+            // NOTE(unsafe) read with no side effects. The data returned will
+            // remain valid for its lifetime because we take an immutable
+            // reference to this FlashWriter, and any operation that would
+            // invalidate the data returned would first require taking a mutable
+            // reference to this FlashWriter.
+            unsafe { core::slice::from_raw_parts::<'static, u8>(address, length) },
```
thalesfragoso, line 217: This shouldn't be `'static`, it should have the same lifetime of `&self`. Not sure if you need to strictly annotate it in order for the compiler to be happy.

src/flash.rs
thalesfragoso: The `'a` in `FlashWriter<'a>` is the lifetime of the borrow of `Parts`, we wouldn't need the `FlashReaderGuard` to borrow the `FlashWriter` for this same lifetime, so having another lifetime parameter helps with that, not sure if that would create real implications or the compiler would be smart enough to solve it.  Either way, it would be better because it wouldn't over constrain our implementation, but anyway, we probably won't need `FlashReaderGuard`...

thalesfragoso:
> So, I've been playing around with different ways to deal with the locking/unlocking of the flash, and I've found things I really don't like about each solution, I'd appreciate some input.

I think we should hold on this change, better to leave for another PR, since the path isn't clear yet and the way it's now isn't that bad.

src/flash.rs
```diff
@@ -2,6 +2,294 @@
...
+        {
+            self.page_erase(offset)?;
+        }
+
+        // Report Success
+        Ok(())
+    }
+
+    /// Retrieve an slice of data from FLASH_START +offset
+    pub fn read(&self, offset: u32, length: usize) -> Result<&[u8]> {
```
ryan-summers, line 202: So, one downside of this approach is that you have to actually copy flash contents into a user buffer, which isn't actually necessary. An alternative API is as follows:

```
struct MemoryGrant {
    writer: FlashWriter,
    data: &'static [u8],
}

struct FlashWriter {};

impl MemoryGrant {
    pub fn return(self) -> FlashWriter;
}

impl FlashWriter {
    pub fn read(self, address, length) -> MemoryGrant;
}
```

This lets you provide a direct reference to the underlying flash and avoids the copy overhead. The call consumes `self` and stores it in the memory grant. You can then impl `Deref` for the `MemoryGrant` to allow users to directly access the underlying memory I believe.

thalesfragoso: Why do you think we're copying the flash contents ? We're creating a slice directly from flash memory, I think...

ryan-summers: If that's the case, can't this lead to unsafe behavior? If someone reads from a flash address, but then writes into the data contained there, wouldn't that result in UB? A potentially immutable object contents would be mutated

irwineffect: As @thalesfragoso mentions [here](https://github.com/stm32-rs/stm32f1xx-hal/pull/257?fbclid=IwAR33994-VoyGLpFfgZuPe38brCxop-6UXKllIu3RQsnxIvWYo5ArB5u0_eQ#issuecomment-681068488), we can't use the FlashWriter while that slice exists.

thalesfragoso: > we can't use the FlashWriter while that slice exists (the borrow checker will prevent it).

Right now it won't because the slice is marked as `'static`, we need to change that.

src/flash.rs
irwineffect: I see, that makes sense. Thanks!

src/flash.rs
irwineffect: Although the unsafe block marked the lifetime as static, but my testing indicated the borrow check was still blocking mutable borrows of the `FlashWriter` based on the slice returned by `read()`. I'm curious why that is? In any case, I've modified to explicitly tie the slice to the lifetime of the immutable borrow in `read()`.

src/flash.rs
irwineffect: It looks like the compiler can infer the lifetime, as this compiles:
```rust
 unsafe { core::slice::from_raw_parts(address, length) }
```
However, I decided to explicitly annotate the lifetimes around `read()` to make the relationship more obvious.

irwineffect:
> > So, I've been playing around with different ways to deal with the locking/unlocking of the flash, and I've found things I really don't like about each solution, I'd appreciate some input.
> 
> I think we should hold on this change, better to leave for another PR, since the path isn't clear yet and the way it's now isn't that bad.

@TheZoq2, if you agree, we can leave reworking the unlock/locking logic for a future PR?

src/flash.rs
thalesfragoso: >  I'm curious why that is?

Hmm, I think I know what was going on, by the lifetime elisions rules, the returned slice would have the same lifetime as the `self` argument, so although we were creating a `'static` slice, its lifetime would shrink to the lifetime of the return type.

src/flash.rs
thalesfragoso: Hmm, but I don't think the `'a` lifetime was the best idea, since it's the same lifetime as `Parts` like in the other case.
To be honest I think we should just let the compiler do the job and use just 
```rust
pub fn read(&self, offset: u32, length: usize) -> Result<&[u8]> {
    //...
    unsafe { core::slice::from_raw_parts(address, length) }
}
```

The lifetime elisions rules seem to be very clear in this case anyway, so it's best to leave the compiler to choose the appropriate lifetime I think. Sorry for all the nitpicking .-.

TheZoq2:
Sounds like a plan to me!

src/flash.rs
irwineffect: Haha don't be sorry, glad to get feedback! I didn't realize annotating the lifetime would change the behavior, I thought I was just making something that was implicit explicit.

thalesfragoso: It was just because it was the same lifetime you used at
```rust
impl<'a> FlashWriter<'a>
```

irwineffect:
If everything looks good now, should I squash-rebase before we merge?

burrbull:
Yes, please.

src/flash.rs
```diff
@@ -2,6 +2,294 @@
...
+            (start_offset..start_offset + length as u32).step_by(self.sector_sz.kbytes() as usize)
+        {
+            self.page_erase(offset)?;
+        }
+
+        // Report Success
+        Ok(())
+    }
+
+    /// Retrieve an slice of data from FLASH_START +offset
```
TheZoq2, line 201: Nitpick: should be `a slice`. Also, would be nice to have the math in a code block, and a space after the `+`

TheZoq2 (changes requested):
Looks great to me, just 2 things remain: a changelog entry and a tiny issue with one of the doc comments :)

src/flash.rs
```diff
@@ -2,6 +2,294 @@
...
+        } else {
+            if self.verify {
+                // By subtracting 1 from the sector size and masking with
+                // start_offset, we make 'start' point to the beginning of the
+                // page, and similarly 'end' at the end of the page. We do this
+                // because the entire page should have been erased, regardless
+                // of where in the page the given 'start_offset' was.
+                let start = start_offset & !(self.sector_sz.kbytes() as u32 - 1);
+                let end = start + self.sector_sz.kbytes() as u32 - 1;
+                for idx in start..end {
```
thalesfragoso, line 173: Should this range be inclusive ? i.e. `start..=end`

irwineffect: You are absolutely right, good catch. Do you think it should be `start..=end`, or remove the `-1` on the line above when `end` is calculated?

burrbull: remove `-1` is better

burrbull: Wait, I did not see previous line.

burrbull, line 173:
```suggestion
                // By subtracting 1 from the sector size and masking with
                // start_offset, we make 'start' point to the beginning of the
                // page. We do this
                // because the entire page should have been erased, regardless
                // of where in the page the given 'start_offset' was.
                let size = self.sector_sz.kbytes() as u32;
                let start = start_offset & !(size - 1);
                for idx in start..start+size {
```

burrbull: Is this what you want to do?

irwineffect: I like that, I think it is more readable than the previous version.

TheZoq2:
Looks like all standing issues are resolved, thanks everyone who helped out!

Issue #256, opened 2020-08-15
SPI doesn't support 16bit data frames


jonnyando:
It looks like the device spi peripherals are initialised with the DFF bit in CR1 cleared, setting the peripheral to 8bit mode. Additionally, blocking::spi::Write, FullDuplex, and Transfer::Default are all only implemented for `<u8>`. Does it make sense to also implement these functions for a 16 bit data frame mode?

Simply implementing a `<u16>` version of these functions wouldn't work as the peripheral would need to know what mode it is in and ensure CR1 is configured correctly.

I'm still new to rust, so don't know where to start on this. Any ideas on how to add this functionality, or whether it is a good addition or not?

TheZoq2:
I see no reason not to implement them. The reason they're not here yet is simply because noone has done it so far.

As you said, some trickery is required.

The simplest way to do so would be to just duplicate the structs, and replace the dataframe size, but we probably want to reduce code duplication.

I think there may be two ways to make it nicer. One is to make most of the current code generic over the type returned, so you would have `Spi1<..., DataFrame>`. Then we would have to add a few functions for the specific dataframe, probably with a separate trait

```rust
impl SpiInternal for Spi1<..., u8> {
    fn setup_data_frame() {
        ...
    }
    fn read_value() -> u8 {
        ...
    }
}
```
Those functions would then be used by the main impl

Another option is to try and move most of the generic impl into free standing functions, and have some thin impl blocks for the different word lengths.

I'm leaning towards the second being nicer, but I'm not sure.

If you want to have a go at implementing this, let me know, I'll gladly help you out if you get stuck!
```

PR #258, opened 2020-08-17
Spi with 16bit framesize

jonnyando:
This is an attempt to implement [issue #256](https://github.com/stm32-rs/stm32f1xx-hal/issues/256).

This adds a `PhantomData` tag to the `Spi` objects, allowing them to be typed as `u8` or `u16` to indicate their data frame size. The existing constructor creates a `Spi<..., u8>` object with no change in syntax to not break the existing API. To convert between dataframe modes you must call `frame_size_8bit()` or `frame_size_16bit()` which consumes the old type and hands back the new one.

Example usage:
```
    let spi8 = spi::Spi::spi2(dp.SPI2, (sclk, miso, mosi), spi_mode, 100.khz(), clocks, &mut rcc.apb1);
    let spi16 = spi.frame_size_16bit();
```

So far there is no way to initialise directly into 16bit data frame mode. I couldn't think of an elegant way without breaking the current API, and didn't like the idea of just having `spi::Spi::spi2_16b_mode(dp.SPI1...`, so I'd love suggestions.

I'd love some feedback on my implementation as I am quite new to rust!

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -231,7 +231,7 @@ pub type SpiRegisterBlock = crate::pac::spi1::RegisterBlock;
 pub trait SpiReadWrite<T> {
     fn read_data_reg(&mut self) -> T;
     fn write_data_reg(&mut self, data: T);
-    fn write_fast(&mut self, words: &[T]) -> Result<(), Error>;
+    fn spi_write(&mut self, words: &[T]) -> Result<(), Error>;
 }
 
 impl<SPI, REMAP, PINS, FrameSize> SpiReadWrite<FrameSize> for Spi<SPI, REMAP, PINS, FrameSize>
@@ -254,7 +254,7 @@ where
     // of RM0008 Rev 20. This is more than twice as fast as the
     // default Write<> implementation (which reads and drops each
     // received value)
-    fn write_fast(&mut self, words: &[FrameSize]) -> Result<(), Error> {
+    fn spi_write(&mut self, words: &[FrameSize]) -> Result<(), Error> {
         // Write each word when the tx buffer is empty
         for word in words {
             loop {
@@ -481,7 +481,7 @@ where
     // default Write<> implementation (which reads and drops each
     // received value)
     fn write(&mut self, words: &[u8]) -> Result<(), Error> {
-        self.write_fast(words)
+        self.spi_write(words)
     }
 }
 
@@ -492,7 +492,7 @@ where
     type Error = Error;
 
     fn write(&mut self, words: &[u16]) -> Result<(), Error> {
-        self.write_fast(words)
+        self.spi_write(words)
     }
 }
 
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -414,6 +414,18 @@ where
         }
     }
 
+    pub fn frame_size_8bit(self) -> Spi<SPI, REMAP, PINS, u8>{
+        self.spi.cr1.modify(|_, w| w.spe().clear_bit());
+        self.spi.cr1.modify(|_, w| w.dff().clear_bit());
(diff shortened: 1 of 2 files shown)


jonnyando:
I've separated out the duplicated code into generic `impl`s where I thought it made sense. The `SpiReadWrite` trait is really only required to still implement the custom `::blocking::spi::Write` implementation from master, otherwise it could be made generic. If I make it `::blocking::spi::Write<FrameSize>`, it conflicts the the generic definition already in `embedded-hal` which I'm not sure how to override (if you even can).

I haven't tested it on hardware yet (I have a STM32F103RB-NUCLEO to test the spi ports in a loopback) so will do that this week.

I also haven't touched DMA yet, so currently the DMA functionality is only implemented for the 8bit dataframe mode. I'm not sure if that should be its own PR.

Also, I saw this note in the code comments:
``` rust
// NOTE(read_volatile) read only 1 byte (the svd2rust API only allows
        // reading a half-word)
        return unsafe { ptr::read_volatile(&self.spi.dr as *const _ as *const u8) };
```
However I haven't been able to verify that myself. Reading the dr `as *const u16` appeared to work in earlier testing the other week, so I'm not sure if this note is true.

I'll add to the changelog once I've had a chance to test it on hardware.

src/spi.rs
```diff
@@ -221,13 +222,94 @@ impl<REMAP, PINS> Spi<SPI3, REMAP, PINS> {
...
+    fn write_data_reg(&mut self, data: FrameSize) {
+        // NOTE(write_volatile) see note above
+        unsafe { ptr::write_volatile(&self.spi.dr as *const _ as *mut FrameSize, data) }
+    }
+
+    // Implement write as per the "Transmit only procedure" page 712
+    // of RM0008 Rev 20. This is more than twice as fast as the
+    // default Write<> implementation (which reads and drops each
+    // received value)
+    fn write_fast(&mut self, words: &[FrameSize]) -> Result<(), Error> {
```
TheZoq2, line 257: Since we don't have a write_slow or similar, I think we could rename this to just `write` as it is currently, and remove the point in the comment about this being faster than something else.

TheZoq2 (commented):
Looks good to me, nice job! Let me know when you've tested on hardware

TheZoq2:
> Also, I saw this note in the code comments:

From what I can tell, that's not relevant anymore. I pressume this used the `stm32f1` crate to read the register in the past, but at some point we moved over to the pointer read which has no such imitations and someone just forgot to remove the comment :)

> I also haven't touched DMA yet, so currently the DMA functionality is only implemented for the 8bit dataframe mode. I'm not sure if that should be its own PR.

I think we can leave it as a separate PR, better to have one feature now and one later than none now ;)

We should make sure you can't create a DMA version that gives `u8` from the `u16` version though

src/spi.rs
jonnyando: I was thinking to differentiate it from the `embedded_hal::blocking::spi::Write<T>` function `fn write(&mut self, words: &[T])`, which calls `write_fast()`. There is a naming conflict if they're both named `write`. Perhaps `spi_write` or something similar would be more appropriate?

jonnyando:
Tested on my Nucleo and successfully sent and received data in a loopback in both 8 and 16 bit modes. I think this is good to go.

TheZoq2:
Awesome! I think there is one small problem left: as far as I understand, this is a breaking change, so a small edit to the changelog is required :)

src/spi.rs
TheZoq2: Oh right, that's a good point. I think spi_write might be a bit clearer.

jonnyando:
That should do it :)

TheZoq2:
Lovely, thanks!

PR #260, opened 2020-08-19
Fix small mistake in time.rs: kilohertz should be megahertz

codemaster97:
Fix error in description of Megahertz


TheZoq2:
Perfect, thanks!

PR #264, opened 2020-09-16
Add TX DMA implementation for SPI3

### LINKED ISSUE #263: DMA traits implemented for `SpiTxDma` for SPIs 1 and 2, but not 3
We're running a large number of LEDs on SPI3 but are running into some performance issues. To increase efficiency we're looking at using the DMA, however it looks like the necessary DMA transfer traits are only implemented for SPIs 1 and 2.

Under the SPI section the stm32f105-7 datasheet mentions:

> All SPIs can be served by the DMA controller.

Is there any reason in particular SPI3's implementations have been omitted? Or perhaps just oversight? The only difference with SPI3 that comes to mind is the necessary AFIO remap, but I'm not experienced enough to know if this affects the DMA at all?

Just thought I'd leave an issue before heading home. If it looks like this should be added, I'll have a go at it once I'm back at the lab tomorrow.

mitchmindtree:
DMA2 channel 2 was selected after referring to the "DMA request mapping"
section of the reference manual (section 13.3.7).

Tested locally with a large number of LEDs and works well.

Closes #263.


src/spi.rs
```diff
@@ -41,6 +41,7 @@ use crate::pac::{SPI1, SPI2};
 
 use crate::afio::MAPR;
 use crate::dma::dma1::{C3, C5};
+use crate::dma::dma2::C2;
```
thalesfragoso, line 44: Could you also `cfg` this to avoid the warnings when compiling for other devices ?

TheZoq2:
Looks good, thanks!

PR #265, opened 2020-09-16
Add missing SPI3 remap

mitchmindtree:
This is necessary in order for SPI3 to function correctly.

Note that this is a breaking change, as it requires access to `MAPR` via
a new argument. This has been added as the third argument to match the
`spi1` constructor.

Tested successfully on an F107.


TheZoq2:
Hmm, this is interesting. Is the spi3 completely unusable without remap?

mitchmindtree:
From memory yes, this allowed me to get it working at all, but I haven't had a chance to triple check this since your comment just yet.

TheZoq2:
Oh, I looked into this in more detail and it makes a lot of sense that this would be needed now. Would you mind fixing the merge conflict in CHANGELOG.md?

TheZoq2:
Thanks!

PR #274, opened 2020-10-03
Change DMA API to use embedded-dma traits

### LINKED ISSUE #270: dma::WriteDma::write is unsound.
The trait `Static` doesn't provide the invariants necessary for a buffer to be safe to use with DMA, it actually doesn't provide any invariant at all, since it's public and safe to implement.

The quickest solution would be to mark the trait as `unsafe` and document it. But a better solution would be to move this API to use the [embedded-dma](https://crates.io/crates/embedded-dma) traits.

This is all breaking changes and we should also discuss if we want to yank previous versions with this problem.

thalesfragoso:
Fixes #270 

This PR is pretty much done, but it would be nice to test at least the affected examples on hardware. Unfortunately, I don't have the time to do it right now, I would appreciate some help.

CC @TheZoq2 and @mitchmindtree

### DIFF
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -317,6 +317,11 @@ macro_rules! dma {
                             // we need a fence here for the same reason we need one in `Transfer.wait`
                             atomic::compiler_fence(Ordering::Acquire);
 
+                            // `Transfer` needs to have a `Drop` implementation, because we accept
+                            // managed buffers that can free their memory on drop. Because of that
+                            // we can't move out of the `Transfer`'s fields, so we use `ptr::read`
+                            // and `mem::forget`.
+                            //
                             // NOTE(unsafe) There is no panic branch between getting the resources
                             // and forgetting `self`.
                             unsafe {
@@ -350,6 +355,11 @@ macro_rules! dma {
                             // we need a fence here for the same reason we need one in `Transfer.wait`
                             atomic::compiler_fence(Ordering::Acquire);
 
+                            // `Transfer` needs to have a `Drop` implementation, because we accept
+                            // managed buffers that can free their memory on drop. Because of that
+                            // we can't move out of the `Transfer`'s fields, so we use `ptr::read`
+                            // and `mem::forget`.
+                            //
                             // NOTE(unsafe) There is no panic branch between getting the resources
                             // and forgetting `self`.
                             unsafe {
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -361,7 +361,7 @@ macro_rules! dma {
                         }
                     }
 
-                    impl<BUFFER, PAYLOAD> Transfer<W, &'static mut BUFFER, RxDma<PAYLOAD, $CX>>
+                    impl<BUFFER, PAYLOAD> Transfer<W, BUFFER, RxDma<PAYLOAD, $CX>>
                     where
                         RxDma<PAYLOAD, $CX>: TransferPayload,
                     {
--- a/src/dma.rs
+++ b/src/dma.rs
@@ -1,16 +1,18 @@
 //! # Direct Memory Access
 #![allow(dead_code)]
 
-use core::marker::PhantomData;
-use core::ops;
+use core::{
+    marker::PhantomData,
+    sync::atomic::{compiler_fence, Ordering},
+};
+use embedded_dma::{StaticReadBuffer, StaticWriteBuffer};
 
 use crate::rcc::AHB;
 
 #[derive(Debug)]
+#[non_exhaustive]
 pub enum Error {
     Overrun,
-    #[doc(hidden)]
-    _Extensible,
 }
 
 pub enum Event {
@@ -33,29 +35,17 @@ where
     readable_half: Half,
 }
 
-impl<BUFFER, PAYLOAD> CircBuffer<BUFFER, PAYLOAD> {
+impl<BUFFER, PAYLOAD> CircBuffer<BUFFER, PAYLOAD>
+where
+    &'static mut [BUFFER; 2]: StaticWriteBuffer,
+    BUFFER: 'static,
+{
     pub(crate) fn new(buf: &'static mut [BUFFER; 2], payload: PAYLOAD) -> Self {
         CircBuffer {
             buffer: buf,
             payload,
             readable_half: Half::Second,
         }
-    }
-}
-
-pub trait Static<B> {
-    fn borrow(&self) -> &B;
-}
-
-impl<B> Static<B> for &'static B {
-    fn borrow(&self) -> &B {
-        *self
-    }
-}
-
-impl<B> Static<B> for &'static mut B {
-    fn borrow(&self) -> &B {
-        *self
     }
 }
 
@@ -70,13 +60,19 @@ pub trait TransferPayload {
     fn stop(&mut self);
 }
 
-pub struct Transfer<MODE, BUFFER, PAYLOAD> {
+pub struct Transfer<MODE, BUFFER, PAYLOAD>
+where
+    PAYLOAD: TransferPayload,
+{
     _mode: PhantomData<MODE>,
     buffer: BUFFER,
     payload: PAYLOAD,
 }
 
-impl<BUFFER, PAYLOAD> Transfer<R, BUFFER, PAYLOAD> {
+impl<BUFFER, PAYLOAD> Transfer<R, BUFFER, PAYLOAD>
+where
+    PAYLOAD: TransferPayload,
+{
     pub(crate) fn r(buffer: BUFFER, payload: PAYLOAD) -> Self {
         Transfer {
             _mode: PhantomData,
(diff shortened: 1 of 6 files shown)


TheZoq2:
Awesome, i'll have a look at it tomorrow :)

src/dma.rs
```diff
@@ -316,7 +317,14 @@ macro_rules! dma {
...
 
-                            (self.buffer, self.payload)
+                            // NOTE(unsafe) There is no panic branch between getting the resources
+                            // and forgetting `self`.
+                            unsafe {
+                                let buffer = ptr::read(&self.buffer);
+                                let payload = ptr::read(&self.payload);
+                                mem::forget(self);
+                                (buffer, payload)
+                            }
```
TheZoq2, line 327: I'm probably missing something here, but why is this required? It seems like this would be the smae as just moving `self.buffer` and `self.payload` into the tuple directly?

src/dma.rs
```diff
@@ -480,27 +495,30 @@ pub trait Transmit {
...
     fn circ_read(self, buffer: &'static mut [B; 2]) -> CircBuffer<B, Self>;
 }
 
+/// Trait for DMA readings from peripheral to memory.
 pub trait ReadDma<B, RS>: Receive
 where
-    B: as_slice::AsMutSlice<Element = RS>,
-    Self: core::marker::Sized,
+    B: StaticWriteBuffer<Word = RS>,
+    Self: core::marker::Sized + TransferPayload,
```
TheZoq2, line 512: Seeing this code, I feel like `TransferPayload` is a missleading name for this trait. Something like "DmaPeripheral" would make more sense to me.

Though we can deal with that later I suppose

src/serial.rs
```diff
@@ -584,15 +585,21 @@ macro_rules! serialdma {
...
+                        // NOTE(unsafe) We own the buffer now and we won't call other `&mut` on it
+                        // until the end of the transfer.
+                        let (ptr, len) = unsafe { buffer.static_write_buffer() };
                         self.channel.set_peripheral_address(unsafe{ &(*$USARTX::ptr()).dr as *const _ as u32 }, false);
-                        self.channel.set_memory_address(buffer.as_ptr() as u32, true);
-                        self.channel.set_transfer_length(buffer.len() * 2);
+                        self.channel.set_memory_address(ptr as u32, true);
+                        self.channel.set_transfer_length(len);
 
                         atomic::compiler_fence(Ordering::Release);
```
TheZoq2, line 605: Does opening a new scope for this still serve a purpose?

src/spi.rs
```diff
@@ -53,20 +53,18 @@ use crate::rcc::{Clocks, Enable, GetBusFreq, Reset, APB1, APB2};
 use crate::time::Hertz;
 
 use core::sync::atomic::{self, Ordering};
-
-use as_slice::AsSlice;
+use embedded_dma::StaticReadBuffer;
 
 /// SPI error
 #[derive(Debug)]
+#[non_exhaustive]
```
TheZoq2, line 60: Are we expecting to add more errors here in the future? IIRC we handle all hardware errors that can occur already.

TheZoq2 (commented):
Looked through the code and added a few nitpicks/questions.

I'll do some testing of the examples now

TheZoq2:
Alright, tested the serial TX and spi TX examples and they still work. The rx examples are a bit more tricky, but I'll try to get something working

src/dma.rs
thalesfragoso: `Transfer` now implements `Drop`, so we can't move out of its fields.

src/dma.rs
thalesfragoso: Today I think we only put peripherals in there, yes. However, the generic name is good in case someone needs to put more stuff in there, but yeah, maybe we can get a better name than `..Payload`.

src/serial.rs
thalesfragoso: It limits the `ptr` scope, but not really necessary, it was there before so I didn't touch it.

src/spi.rs
thalesfragoso: I think it's good practice to avoid breaking changes, but again, it was there before (implemented manually), I just changed to please clippy. I would leave this decision to another PR, I prefer not to change behavior not related to DMA in this one.

src/dma.rs
TheZoq2: Right, the genericity is a good point. I can't really think of a good name then :/

Maybe something like TransferSource, though that doesn't work when it's the target of the transfer

src/serial.rs
TheZoq2: Yea, that's what I figured. I'd say it's better to remove the scope then, just to avoid people having to ask that question every time they look at the code :D

src/spi.rs
TheZoq2: Oh right, I missed the _Exclusive thing. Yea, let's leave that to the future

src/dma.rs
TheZoq2: Ah right. Could you add a comment clarifying that in the code? Seems like something that will come up in the future

thalesfragoso:
@TheZoq2 Alright, I'm marking this as ready for review, but if you could test the circular adc example (adding some prints to check it), it would be great.

TheZoq2:
Finally got around to testing the circular ADC DMA. Looks like things work as expected :tada:

TheZoq2:
I guess the question that remains is what we should do with the previous versions.

Given the severity, it might make sense to yank the affected version. According to crates.io, we still have plenty of people on 0.6, 0.5 and older versions, and we have quite a few breaking changes lined up.

Would it make sense to yank the latest versions of 0.4, 0.5 and 0.6, then release a new version of them with the DMA things marked as deprecated? That way people who don't use DMA can avoid upgrading right now, while people who do get notified about the problem.

therealprof:
Hm, the yanking mechanics is subtle. I'm not sure it would have the intended effect if we did that. But I'll leave that decision up to you.

TheZoq2:
In what way is it subtle. I have not dealt much with yanking so I may be missing something here.

As I understand it the effect will be that people who clean build their project will either get bumped to the patched version and see the deprecation warning, or get a message about the version they are using (in the lockfile or Cargo.toml) being yanked. Obviously we can't do much for people who don't do a clean build

thalesfragoso:
> then release a new version of them with the DMA things marked as deprecated

I don't think deprecation conveys the right idea here, so not sure if it's worth it, and I guess people can still use yanked versions without problems (if they want to), no ?

therealprof:
> In what way is it subtle. I have not dealt much with yanking so I may be missing something here.

Me neither but the exact workings are a mystery to pretty much everyone which is why it is subtle.

> As I understand it the effect will be that people who clean build their project will either get bumped to the patched version and see the deprecation warning, or get a message about the version they are using (in the lockfile or Cargo.toml) being yanked. Obviously we can't do much for people who don't do a clean build

My understanding is that it pretty much only works for clean builds without Cargo.lock. If you have a Cargo.lock it'll continue working, if you don't then I think it depends on what the and previous versions are and what is specified in the Cargo.toml: If you have a minor specified and there's a minor update it'll fall forward, if the minor version is the last one in row (which is typical if you have breaking changes due to <yanking reason>) it'll fall back to the previous patch version (which typically has the same problem). To upgrade the minor you need to change the Cargo.toml.

TheZoq2:
> I don't think deprecation conveys the right idea here, so not sure if it's worth it, and I guess people can still use yanked versions without problems (if they want to), no ?

I'm not sure, I thought yanked versions are not distributed anymore? As for conveying the right idea, I figure a correctly worded deprecation message should give roughly the right idea.

> My understanding is that it pretty much only works for clean builds without Cargo.lock.

Yea, i'm not sure how cargo.lock will affect things with yanked versions.

> and there's a minor update it'll fall forward, if the minor version is the last one in row (which is typical if you have breaking changes due to ) it'll fall back to the previous patch version

This will not be a problem if we release this new version with deprecated DMA though. 

> To upgrade the minor you need to change the Cargo.toml.
I don't believe so, unless you have explicitly requested a specific version. The default of
`x.y.z` actually `x.y.z^` as I understand it.

therealprof:
> `x.y.z` actually `x.y.z^` as I understand it.

You probably mean `^x.y.z`. 

But minor level means `y`, not `z` which is the patch level. Below 1.0 the Rust compatibility rules say that only the patch level may automatically change to a higher version. Unfortunately tools like `cargo` add always add a patch version so you can't even fall back to a previous patch version (if such a version would exist, that is).

TheZoq2:
Oh right, I may be getting things confused with the patch vs minor version thing then since I'm so used to 0.x crates

What I'm suggesting is that we release a 

0.7.0: with fixed DMA
0.6.2: 0.6.1 with DMA deprecated
0.5.3: 0.5.2 with DMA deprecated

As I understand it, that would automatically bump clean builds of 0.6.x and 0.5.x users to a non-breaking version with a deprecation message, while new users would get the 0.7 version

TheZoq2:
Alright, 0.7 has been released :tada:

PR #277, opened 2020-10-15
Dma Transfer

burrbull:
As we now use `StableDeref` from `embedded-dma`, shall DMA still take owned values of buffer and peripheral?
Or we can change API to take `& mut PAYLOAD`?

### DIFF
--- a/examples/adc-dma-rx.rs
+++ b/examples/adc-dma-rx.rs
@@ -35,14 +35,14 @@ fn main() -> ! {
     // Configure pa0 as an analog input
     let adc_ch0 = gpioa.pa0.into_analog(&mut gpioa.crl);
 
-    let adc_dma = adc1.with_dma(adc_ch0, dma_ch1);
-    let buf = singleton!(: [u16; 8] = [0; 8]).unwrap();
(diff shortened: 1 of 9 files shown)


TheZoq2:
This seems like a good change, unless there are any soundness issues.

CC: @thalesfragoso

thalesfragoso:
This is unsound, unfortunately.

`StableDeref` by itself isn't enough to make this API safe, the user can still safely `mem::forget` the `Transfer` and then re-use the buffer (or pop its stack) before the transfer is finished, causing memory corruption.

TheZoq2:
Alright, glad I asked :). In that case I'll go ahead and close this

burrbull:
> the user can still safely `mem::forget` the `Transfer` and then re-use the buffer (or pop its stack) before the transfer is finished, causing memory corruption

If that's the only problem, maybe it makes sence just to add `blocking_read()`/`blocking_write()` (or `read_and_wait()`/`write_and_wait()`) functions to DmaRead/DmaWrite traits, that take `&mut self`, run transfer and wait complete instead of return `Transfer` struct?

TheZoq2:
Oh yea, that might be a good idea for when you want high speed transfer but don't care about asyncronous transfers

PR #278, opened 2020-10-17
Add BlockingReadDma/BlockingWriteDma traits

burrbull:
cc @TheZoq2 
@thalesfragoso

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -737,14 +737,12 @@ where
     }
 }
 
-impl<B, PINS, MODE> crate::dma::ReadDma<B, u16> for AdcDma<PINS, MODE>
-where
-    Self: TransferPayload,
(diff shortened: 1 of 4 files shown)


thalesfragoso:
Thanks @burrbull , that indeed seems like a good addition, but maybe we could use a closure approach, like crossbeam's scoped threads and the `scoped_threadpool` crate. That way the user can still do some work while the DMA is running, but I'm not sure how we would handle panics in the closure, it should be okay since we don't usually unwind, but still not nice, I have to think more about it.

Another thing, `WriteBuffer` and the other traits are implemented for pointer types, so `B` is already a pointer, you don't need `&mut B`, changing `StaticWriteBuffer` to `WriteBuffer` should work better.

PS: I haven't reviewed all of this PR yet.

PR #280, opened 2020-10-26
Add SPI frame format config method

### LINKED ISSUE #279: SPI no frame format option.
Hi, 
When i was writing driver to some device i've found that HAL implementation does not allow me to set frame format. In my case this is lsbfirst which i have to choose. Can you explain me how can i change it from default msbfirst ?

repnop:
Fixes #279

This adds a method to the SPI type which allows configuring which frame format to use for data transfers.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -116,6 +116,15 @@ pub struct Spi<SPI, REMAP, PINS, FRAMESIZE> {
     _framesize: PhantomData<FRAMESIZE>,
 }
 
+/// The bit format to send the data in
+#[derive(Debug, Clone, Copy)]
+pub enum SpiBitFormat {
+    /// Least significant bit first
+    LsbFirst,
+    /// Most significant bit first
+    MsbFirst,
+}
+
 /// A filler type for when the SCK pin is unnecessary
 pub struct NoSck;
 /// A filler type for when the Miso pin is unnecessary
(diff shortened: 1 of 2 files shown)


src/spi.rs
```diff
@@ -116,6 +116,15 @@ pub struct Spi<SPI, REMAP, PINS, FRAMESIZE> {
...
 
+/// The frame format to send the data in
+#[derive(Debug, Clone, Copy)]
+pub enum SpiFrameFormat {
+    /// Least significant byte first
+    Lsb,
+    /// Most significant byte first
+    Msb,
+}
+
```
TheZoq2, line 127: These comments should say `Most significant bit first` I believe.

Also, I almost want to call the members `[ML]sbFirst` but that becomes very long.

TheZoq2 (changes requested):
Thanks for the PR, looks good apart from a few naming and doc issues.

From reading the reference manual, it looks like the term "frame format" refers to both bit order and word length. Would `SpiBitOrder` be a better name?

src/spi.rs
repnop: yeah naming these is kind of tricky because if you want to spell it out so its obvious it becomes super long, but otherwise its kind of awkward. perhaps `MostSigBitFirst`|`LeastSigBitFirst`? "sig" feels awkward but its Close Enough:tm: I think, thoughts?

TheZoq2: I think LSB and MSB are well known enough that we can call them `LsbFirst` and `MsbFirst`

TheZoq2 (approved):
LGTM!

PR #281, opened 2020-11-02
Copy memory.x to output folder

jaysonsantos:
This makes projects like defmt work out of the box.

Is this worth it to add to the changelog?

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,17 @@
+use std::env;
+use std::fs::File;
+use std::io::Write;
+use std::path::PathBuf;
+
+fn main() {
+    // Put the linker script somewhere the linker can find it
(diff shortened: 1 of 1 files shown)


thalesfragoso:
This crate has support for several different microcontrollers with different flash and ram sizes, so I'm not sure if this is a good idea.

jaysonsantos:
Hey there, what if the `build.rs` chooses the memory.x for the specific feature/device e.g. stm32f100?

TheZoq2:
Unfortunately, there is not a fixed amount of ram for each chip family.

For example, the f103 has everything from 16k/6 to 1M/96k ram and flash

image

https://www.st.com/en/microcontrollers-microprocessors/stm32f103.html

jaysonsantos:
Hey there, that makes sense.
I guess I will do it on my own build.rs.
Thanks!

PR #282, opened 2020-11-03
various i2c fixes

sjoerdsimons:
Fixes required to work with an AM2320 sensor; This sensor doesn't acknowledge the first i2c message sent to it, but it triggers it to wake up. This caused the i2c state to stay in an error state and the bus not to go idle.  This is fixed in the first two patches

Later on a multi (>2) byte read is necessary  to actually get the data; However the data didn't get properly acknowledged, fixed in the last patch.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -533,6 +533,7 @@ where
                 self.nb.i2c.cr1.modify(|_, w| w.ack().set_bit());
             }
             buffer_len => {
+                self.nb.i2c.cr1.modify(|_, w| w.ack().set_bit());
                 self.nb.i2c.sr1.read();
                 self.nb.i2c.sr2.read();
 
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -436,12 +436,17 @@ where
         last_ret
     }
 
-    fn write_without_stop(&mut self, addr: u8, bytes: &[u8]) -> NbResult<(), Error> {
-        self.send_start_and_wait()?;
+    fn send_addr_and_wait(&mut self, addr: u8, read: bool) -> NbResult<(), Error> {
         self.nb.i2c.sr1.read();
-        self.nb.send_addr(addr, false);
+        self.nb.send_addr(addr, read);
+        let ret = busy_wait_cycles!(wait_for_flag!(self.nb.i2c, addr), self.addr_timeout);
+        if ret == Err(Other(Error::Acknowledge)) {
+            self.nb.send_stop();
+        }
+        ret
+    }
 
-        busy_wait_cycles!(wait_for_flag!(self.nb.i2c, addr), self.addr_timeout)?;
+    fn write_bytes_and_wait(&mut self, bytes: &[u8]) -> NbResult<(), Error> {
         self.nb.i2c.sr1.read();
         self.nb.i2c.sr2.read();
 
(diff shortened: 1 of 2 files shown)


src/i2c.rs
```diff
@@ -436,12 +436,17 @@ where
         last_ret
     }
 
-    fn write_without_stop(&mut self, addr: u8, bytes: &[u8]) -> NbResult<(), Error> {
-        self.send_start_and_wait()?;
+    fn send_addr_and_wait(&mut self, addr: u8, read: bool) -> NbResult<(), Error> {
         self.nb.i2c.sr1.read();
```
TheZoq2, line 440: Not directly related to the PR, but I'm now becomming unsure about what this is doing here. My best guess is that it tries to clear the status registers, but clearly that's not done since the first commit in the PR is needed...

TheZoq2 (commented):
Thanks for the PR, that sounds like it must have been an annoying bug to track down :sweat_smile: 

I added a thought about the random status register reads which I don't know the purpose of.

Would you mind adding an entry to the changelog about this? Other than that, it looks great!

src/i2c.rs
sjoerdsimons: hmm; looking at the datasheet i'm not entirely sure if it's needed here either; Fwiw the relevant parts are:
```
Then the master waits for a read of the SR1 register followed by a write in the DR register
with the Slave address (see Figure 273 and Figure 274 Transfer sequencing EV5).
```
with EV5 being `EV5: SB=1, cleared by reading SR1 register followed by writing DR register with Address.`

But ofcourse the `send_start_and_wait` is already reading SR1 while polling for  the flag; So it's probably not needed here.

Overall it seems some bits of sr1 are persistents (the error bits in particular) while others gets cleared on read

TheZoq2: Right. I think the best course of action here is to break this into its own issue and maybe fix it later since doing the wrong thing could break things subtly.

freiguy1:
Thanks for this fix! I was beating my head against the wall trying to figure out why I couldn't use a library to read from a real time clock module. It was panicking when trying to get the date & time because it read 8 bytes through I2C.

src/i2c.rs
```diff
@@ -220,12 +220,16 @@ macro_rules! wait_for_flag {
...
+            $i2c.sr1.write(|w| w.berr().clear_bit());
             Err(Other(Error::Bus))
         } else if sr1.arlo().bit_is_set() {
+            $i2c.sr1.write(|w| w.arlo().clear_bit());
             Err(Other(Error::Arbitration))
         } else if sr1.af().bit_is_set() {
+            $i2c.sr1.write(|w| w.af().clear_bit());
             Err(Other(Error::Acknowledge))
         } else if sr1.ovr().bit_is_set() {
+            $i2c.sr1.write(|w| w.ovr().clear_bit());
```
Disasm, line 232: Hmm, is `write` correct here? These look like no-op to me.

TheZoq2: I could be wrong, but I believe these are there to reset the flags

Disasm: Yes, but `sr1.write(|w| w.any_field().clear_bit());` basically writes 0 to the whole register, regardless of the field name. That's why this looks like a mistake.

TheZoq2: Oh right, I guess that should be `modify` instead. Good catch, I'll go ahead and fix that

PR #283, opened 2020-11-09
add RNG

rise0chen:
Thanks for contributing to the project!

Please write a short description of your changes.

We also maintain a changelog in CHANGELOG.md. If you add a feature or fix a bug, please add an entry there :)

### DIFF
--- a/examples/rng.rs
+++ b/examples/rng.rs
@@ -1,0 +1,24 @@
+#![no_main]
+#![no_std]
+
+use cortex_m_rt::entry;
+use cortex_m_semihosting::hprintln;
+use panic_halt as _;
+use stm32f1xx_hal::{adc, pac, prelude::*};
+
+#[entry]
+fn main() -> ! {
+    let dp = pac::Peripherals::take().unwrap();
+    let mut flash = dp.FLASH.constrain();
+    let mut rcc = dp.RCC.constrain();
+    let clocks = rcc.cfgr.freeze(&mut flash.acr);
+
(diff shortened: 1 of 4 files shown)


thalesfragoso:
Thanks for your contribution, but one thing, I'm really uncomfortable calling this a RNG, this feature is nice and can help a lot of people, but I think we should name it accordingly to avoid confusion.

rise0chen:
It is named by [embedded-hal](https://docs.rs/embedded-hal/0.2.4/embedded_hal/blocking/rng/index.html)

TheZoq2:
I agree with thales here. I think the embedded-hal RNG is supposed to refer to an actual random number generator that has nice properties, perhaps even one that is cryptographically sound. I'll have to look into it a bit more though

TheZoq2:
Finally found some time to look into this. Consensus on what the requirements on RNG trait requirements don't seem to be super clear but given that, I'd say we can use this as long as we label it as what it is.

Jamesmunns also pointed out this: https://github.com/jamesmunns/knurling-test-adapter/blob/main/crates/stm32f4-prng/src/lib.rs which is using a similar method to seed a proper pseudo RNG which should give more "predictable" results.

If nothing else, would you mind adding some documentation to your code describing how this gets its randomness and what properties it has?

rise0chen:
finish

thalesfragoso:
I would still like more public documentation stating the problems with this "RNG" and maybe a type renaming.

burrbull:
`embedded_hal::rng` is deprecated in favour of `rand_core`

PR #291, opened 2020-12-04
Added support for SPI Alternate<OpenDrain> configuration on Mosi and Sck pins

johnnyegel:
This patch allows for OpenDrain configuration on SPI Sck and Mosi pins. This in turn makes it possible to use the SPI interface on 5V bus'es using a simple 5V pull-up (no level shifter).


TheZoq2:
Great, thanks!

Just one small thing, would you mind adding a changelog entry

johnnyegel:
Sorry for taking so long, but I now added a Changelog entry for the change.

TheZoq2:
Thanks! Glad you took that long because it made my delay with merging seem smaller :)

PR #292, opened 2020-12-04
Re-implemented DMA Transferable Trait and added SPI DMA example which use pure abstract references

johnnyegel:
For some reason the Transferable trait has been removed in 0.7.0. This seems to make it impossible to create functions which takes an abstract SPI DMA reference and transmit data via it. A real life use-case for this is in multipurpose drivers, which should work with any SPI peripheral and possibly share it with other drivers.
I've also improved the SPI-DMA example to showcase this exact use case. As a side perk, this will break if pure abstract referencing is ever broken. Thus hopefully insuring this feature is retained in the HAL for the Future.
NOTE! I really think this DMA scheme should be defined on a higher level, and be shared across all the HALs.
Also NOTE! I am very much a noob in Rust (though I have a long experience in MCU/Embedded development in C), so please go easy on me if this is udder crap :)

### DIFF
--- a/examples/spi-dma.rs
+++ b/examples/spi-dma.rs
@@ -9,10 +9,10 @@ use panic_halt as _;
 use cortex_m::singleton;
 use cortex_m_rt::entry;
 use stm32f1xx_hal::{
+    dma::{Transfer, Transferable, WriteDma, R},
     pac,
     prelude::*,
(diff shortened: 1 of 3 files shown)


johnnyegel:
Added changelog entry for change.

PR #296, opened 2020-12-17
Adapt to embedded-hal 1.0.0-alpha.4

### LINKED ISSUE #295: embedded-hal-alpha.4?
Is there a branch or fork of stm32f1xx-hal that builds with embedded-hal-alpha.4? I would like to do some testing with a crate (rust-radio-sx127x) that has moved to that, and the spi setup has to also use that version or the traits don't work properly.

eldruin:
Superseeds #251 
Closes #295


TheZoq2:
Thanks for putting in the work for this. I still think we should wait until e-h leaves alpha before merging, but feel free to convince me otherwise :)

alexmadeathing:
Embedded-hal is now out of alpha! Would love to see this library migrated. I'm happy to provide an extra hand if needed.

burrbull:
Closed. See #416

PR #297, opened 2020-12-17
Use cortex-m-rtic instead of cortex-m-rtfm

eldruin:
Following the upstream rename.


TheZoq2:
Awesome thanks, I forgot we hadn't made that change yet

PR #298, opened 2020-12-23
Use modify instead of write for clearing sr1

TheZoq2:
As per the discussion in #282 

This seems better, but it also leaves the possibility of some flags not being cleared if multiple flags are raised at the same time. Is that something that can even happen, and how should we deal with it then?

CC: @Disasm

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -271,6 +271,9 @@ impl<I2C: Instance> I2c<I2C> {
 
     /// Generate START condition
     fn send_start(&mut self) {
+        // Clear all pending error bits
+        // NOTE(unsafe): Writing 0 clears the r/w bits and has no effect on the r bits
+        self.i2c.sr1().write(|w| unsafe { w.bits(0) });
         self.i2c.cr1().modify(|_, w| w.start().set_bit());
     }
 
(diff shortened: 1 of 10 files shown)


Disasm:
I think that all errors should be additionally cleared before a transaction begins.

burrbull:
I think this code is not optimal. You read sr1 twice. Before comparison and when do modify.

src/i2c.rs
```diff
@@ -220,16 +220,16 @@ macro_rules! wait_for_flag {
         let sr1 = $i2c.sr1.read();
 
         if sr1.berr().bit_is_set() {
-            $i2c.sr1.write(|w| w.berr().clear_bit());
+            $i2c.sr1.modify(|_r, w| w.berr().clear_bit());
```
burrbull, line 223:
```suggestion
            $i2c.sr1.write(|w| w
                 .bits(sr1.bits())
                 .berr().clear_bit()
            );
```
Something like this

Disasm: In theory, one should use an all-one mask instead of the previous register value, because the register value can be changed during the read-modify-write sequence and this code will end up clearing bits it wasn't supposed to clear.

burrbull: I agree
изображение

TheZoq2:
Updated the PR with the suggested changes, status flags should now be cleared before starting a new transaction, and only the error we actually noted is cleared.

I hope the unsafe I had to add is correct, writing to read only registers should have no effect.

We also don't have a way to read more than one error flag, perhaps that's something we should add at some point

burrbull (approved):

PR #299, opened 2020-12-27
[WIP] Implement flash read un/protection per PM0075

TheRealBluesun:
## Abstract

This PR exposes the Read protection feature offered by STMF1xx series MCUs. This feature is implemented as described in the **PM0075 Programming manual** *Doc ID 17863 Rev 2*.

Please do review the comments and formatting -- I'd love some feedback there.

Side note -- this was a very fun first contribution to the Rust embedded community; long overdue.


src/flash.rs
```diff
@@ -101,6 +114,111 @@ impl<'a> FlashWriter<'a> {
...
+
+    fn set_rdp(&mut self, val: u8) -> Result<()> {
+        // First we need to unlock the Option Byte programming
+        self.unlock_opt()?;
+
+        // Wait for operation to finish
+        while self.flash.sr.sr().read().bsy().bit_is_set() {}
+
+        // First we must erase the Option Bytes
+        self.erase_opt()?;
```
TheRealBluesun, line 181: Taking a second look -- this needs to retain the `USER` byte value and others. Maybe we can get away without erasing them entirely? The documentation is a little ambiguous.

TheRealBluesun: Update on this:

Writing the RDP value will fail if we haven't erased the Option Bytes first. I've tried saving off the upper half-word and writing it with the new RDP value, but I can't get it to accept it.  Will have to investigate more.

PR #303, opened 2021-01-10
Use CamelCase for gpio::Edge enum variants

Piroro-hs:
ref: [h7xx](https://github.com/stm32-rs/stm32h7xx-hal/pull/179), [l4xx](https://github.com/stm32-rs/stm32l4xx-hal/pull/187), [f4xx](https://github.com/stm32-rs/stm32f4xx-hal/pull/250), [f7xx](https://github.com/stm32-rs/stm32f7xx-hal/pull/111)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Breaking changes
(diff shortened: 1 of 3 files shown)


therealprof:
CF discussion in https://github.com/stm32-rs/stm32f4xx-hal/pull/250

TheZoq2:
Looks like the rest of the ecosystem are starting to merge this, so I guess we'll follow through. I'll hold off on actually merging it until we have more pressing breaking changes to merge though.

burrbull (approved):

PR #308, opened 2021-02-10
Support embedded_hal::blocking::spi::WriteIter

pozix604:
I needed to write an iterator of `u8` via SPI. My data is in separate arrays but needs to be written with a continuous SPI clock. If I make multiple calls to write instead, the clock would be interrupted in-between each write.

Changes are what I think are minimal to support this functionality.


burrbull:
Add this to changelog, please.

burrbull (approved):

Issue #24, opened 2019-02-17
I2c Example


andrewmruddy:
*New to Rust* - Would it be possible to have an example for I2c? After looking through the documentation, The Rust Book, and other examples, I am still having trouble figuring out proper implementation.

TheZoq2:
Writing an example without using a specific device is difficult.

I will try to write some documentation for the i2c device though, is there anything in particular you're having trouble with?

andrewmruddy:
I have the following for setting up the i2c peripheral: 
`
    let peripherals = stm32f103::Peripherals::take().unwrap();
    let mut gpiob = peripherals.GPIOB.split(&mut rcc.apb2);
    let sda = gpiob.pb7.into_alternate_open_drain(&mut gpiob.crl);
    let scl = gpiob.pb6.into_alternate_open_drain(&mut gpiob.crl);

    let i2c_bus = I2c::i2c1(
        peripherals.I2C1,
        (scl,sda),
        &mut afio.mapr,
        hal::i2c::I2c::Mode::Standard, // < i am having problems with the enum "Mode"
        clocks,
        &mut rcc.apb1,
    );` I am using the MMA8451 accelerometer (adafruit breakout board) for my proof-of-concept.

therealprof:
@andrewmruddy There're some examples for I2C use with `stm32f1xx-hal` in the `ssd1306` crate, e.g. https://github.com/jamwaffles/ssd1306/blob/master/examples/graphics_i2c.rs

andrewmruddy:
Thank you so much!

jounathaen:
If you have a working example, you could post it here.
This would surely help others (like me, I'm planning to do something simmilar :-D )

therealprof:
@jounathaen I wouldn't want to add external dependencies for examples and unfortunately I2C without connected devices is not too impressive/useful.

jounathaen:
@therealprof I can understand that dependencies are undesired, but first I think that you can have I2C examples without external dependencies, like reading generic registers, and second I think that having an outdated example is still better than no up-to-date example.
I was not necessarily thinking about full blown working examples, but code snippets for the documentation would be helpful.
(Of course just my humble opinion)

therealprof:
@jounathaen Oh sure, I definitely agree we should have proper code documentation with logical code snippets.

jaxter184:
[...]

PR #313, opened 2021-02-27
Example of using i2c with the bme280 air sensor

fishrockz:
I created this example for my self as a reference for a more complex use of the bme280 sensor. And thought it might be nice to share.

I noted the comments in #24 that it seems like you don't want examples of real hardware, but I don think this is explicitly stated. Would you like a example like this? I might offer it to the bme280 crate but there example is i2c master agnostic, ill just put it on my blog as reference if this is not the place for it, but i think it will be much more easily discoverable here.

You can get bme280 breakout boards for less than £20 in the uk, eg https://thepihut.com/products/adafruit-bme280-i2c-or-spi-temperature-humidity-pressure-sensor so there pretty cheap and cheerful so it seems like a nice thing to use as a example as its cheap and you can see the values change quite quickly if you put your finger on it or blow on it.

If you would like it how would you like the dep added? the way i have added it, it is only ever built if some one asks for it but it might be nicer as a dev dep, but i don't think they can be optional.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,5 +1,5 @@
 [target.thumbv7m-none-eabi]
-runner = 'gdb-multiarch'
+runner = 'arm-none-eabi-gdb'
 rustflags = [
   "-C", "link-arg=-Tlink.x",
 ]
--- a/examples/i2c-bme280.rs
+++ b/examples/i2c-bme280.rs
@@ -1,0 +1,104 @@
+//! Reads data from a BME280 over i2c
+//!
+//! This assumes that a BME280 is connected with clk on PB6 and data on PB7.
+//!
+//! For the Adafruit breakout boards PB6 should be connected to SCK and PB7 to SDI
+//!
+//! This program writes the sensor values to the debug output provided by semihosting
+//! you must enable semihosting in gdb with `monitor arm semihosting enable` I have it
+//! added to my `.gdbinit`. Then the debug infomation will be printed in your openocd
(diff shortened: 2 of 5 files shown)


burrbull:
Rebase, please.

TheZoq2:
I don't think we should add deps for examples, but I would also like to have a way to have example code like this, perhaps in a separate repo?

eldruin:
If you are interested, I have a lot of examples using this HAL [here](https://github.com/eldruin/driver-examples/tree/master/stm32f1-bluepill)

TheZoq2:
Oh nice, that's a great resource. Would you mind if we make that more official by linking to it in the readme and directing examples like this one there? @eldruin

eldruin:
Sure, I would be happy if it gets more visibility. I have tried to write many self-contained examples for different devices.

fishrockz:
Sorry, I didn't notice the conversation on this. The dep is only acitve if requested as a feature, other repos like Druid support examples that need extras like that.

If this example is useful i can look up adding the dep as part of the example, i think that's possible. i didn't want to put too much effort in to meta stuff like that until I knew if it would be helpful.

Might this be merged if i can sort out the deps?

fishrockz:
@burrbull this is re-based now, if you are interested.

burrbull:
LGTM

TheZoq2:
Oh cool, in the current state, I think this is a great change, thanks!

PR #320, opened 2021-03-12
Update USB driver

Disasm:


examples/usb_serial.rs
```diff
@@ -15,7 +15,7 @@ use usbd_serial::{SerialPort, USB_CLASS_CDC};
 
 #[entry]
 fn main() -> ! {
-    let dp = pac::Peripherals::take().unwrap();
```
TheZoq2, line 18: We have deprecated the `stm32` name in favor of `pac`

src/lib.rs
```diff
@@ -180,10 +180,7 @@ pub mod spi;
 pub mod time;
 #[cfg(feature = "device-selected")]
 pub mod timer;
-#[cfg(all(
-    feature = "stm32-usbd",
-    any(feature = "stm32f102", feature = "stm32f103")
-))]
+#[cfg(any(feature = "stm32f102", feature = "stm32f103"))]
```
TheZoq2, line 183: Coud `feature=stm32-usbd` be used instead here?

src/usb.rs
```diff
@@ -24,11 +24,12 @@ unsafe impl UsbPeripheral for Peripheral {
     const DP_PULL_UP_FEATURE: bool = false;
     const EP_MEMORY: *const () = 0x4000_6000 as _;
     const EP_MEMORY_SIZE: usize = 512;
+    const EP_MEMORY_ACCESS_2X16: bool = false;
 
     fn enable() {
-        let rcc = unsafe { (&*RCC::ptr()) };
+        cortex_m::interrupt::free(|_| unsafe {
```
TheZoq2, line 30: Will this not make the `unsafe` region larger than nessecary? It's only required for the ptr, right?

TheZoq2 (commented):
Thanks for the PR. I left a few notes, but in general, this looks good

src/usb.rs
Disasm: As far as I understand unsafety, the unsafe operation here is accessing the memory via a pointer constructed in unsafe way, not just constructing the pointer. So larger unsafe region looks more correct compared to what we had before.

TeXitoi: No, the unsafe code create a reference, not a pointer. So creating this reference is hightly unsafe, but using the reference is not.

TeXitoi: Creating this reference is better inside the `interrupt::free`. So I propose to just move the `let rcc` line inside the interrupt::free.

TeXitoi (dismissed):

src/usb.rs
burrbull: With #344 be merged we don't need critical section as it uses bit-banding.

burrbull (approved):

PR #322, opened 2021-03-18
Timer additions: raw access to prescaler and auto-reload

Windfisch:
I frequently use timers for precise measurements, and unfortunately, `start_countdown` performs intransparent calculations which can introduce rounding errors.

This PR adds the capability to set and read the raw prescaler and auto-reload values, also we can now read the timer's count; this is useful when ARR is set to `0xFFFF`, i.e. the timer is just "free-running".

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -442,21 +442,8 @@ macro_rules! hal {
                 where
                     T: Into<Hertz>,
                 {
-                    // pause
-                    self.tim.cr1.modify(|_, w| w.cen().clear_bit());
-
                     let (psc, arr) = compute_arr_presc(timeout.into().0, self.clk.0);
-                    self.tim.psc.write(|w| w.psc().bits(psc) );
-
-                    // TODO: Remove this `allow` once this field is made safe for stm32f100
-                    #[allow(unused_unsafe)]
-                    self.tim.arr.write(|w| unsafe { w.arr().bits(arr) });
-
-                    // Trigger an update event to load the prescaler value to the clock
-                    self.reset();
-
-                    // start counter
-                    self.tim.cr1.modify(|_, w| w.cen().set_bit());
+                    self.restart_raw(psc, arr);
                 }
 
                 fn wait(&mut self) -> nb::Result<(), Void> {
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -314,6 +314,14 @@ macro_rules! hal {
                     }
                 )?
 
+                pub fn start_raw(self, psc: u16, arr: u16) -> CountDownTimer<$TIMX>
+                {
+                    let Self { tim, clk } = self;
+                    let mut timer = CountDownTimer { tim, clk };
+                    timer.start_raw(psc, arr);
+                    timer
+                }
+
                 /// Resets timer peripheral
                 #[inline(always)]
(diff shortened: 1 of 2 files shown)


Windfisch:
Hm, seems like it fails to build with `cargo check --features=stm32f100,rt` but works with `cargo check --features=stm32f100,rt`.

I don't understand why it gives me this error for stm32f100, but not for f103.
```
error[E0133]: call to unsafe function is unsafe and requires unsafe function or block
   --> src/timer.rs:366:44
    |
366 |                       self.tim.arr.write(|w| w.arr().bits(arr) );
    |                                              ^^^^^^^^^^^^^^^^^ call to unsafe function
```

I could just wrap around an `unsafe`, but I'd prefer discussing this first. Is this a bug in the stm32 crate or am I missing something? Please help.

TheZoq2:
I think there are some inconsistencies between the different f1 variants which causes that unsafe thingy, we can probably add an `unsafe` block and accept the warning it will throw on the other versions

src/timer.rs
```diff
@@ -347,6 +356,37 @@ macro_rules! hal {
...
                 }
 
+                /// Restarts the timer in count down mode with user-defined prescaler and auto-reload register
+                pub fn restart_raw(&mut self, psc: u16, arr: u16)
+                {
+                    // pause
+                    self.tim.cr1.modify(|_, w| w.cen().clear_bit());
+
+                    self.tim.psc.write(|w| w.psc().bits(psc) );
+                    self.tim.arr.write(|w| w.arr().bits(arr) );
```
TheZoq2, line 366: From the normal `start` function, add `unsafe` and a `#[allow(unused_unsafe)]` attribute to silence the warnings.

src/timer.rs
```diff
@@ -347,6 +356,37 @@ macro_rules! hal {
...
+
+                    // Trigger an update event to load the prescaler value to the clock
+                    self.reset();
+
+                    // start counter
+                    self.tim.cr1.modify(|_, w| w.cen().set_bit());
+                }
```
TheZoq2, line 373: Could you make the old `start` function call this to avoid code duplication

TheZoq2 (changes requested):
Thanks for the PR and sorry for the review delay. I added a few notes which would be nice if they are addressed

Windfisch:
Just to give a update, I've seen your suggestions and will implement them later this week. Thanks for the help :)

Windfisch:
I've implemented your requested changes and merged with the current master. Hopefully everything is fine now :)

burrbull:
> merged with the current master

Don't merge master in PRs, you make commit history non-linear. Rebase only.

Windfisch:
Thanks for the hint, I've updated my PR.

It may be helpful to state "rebase over merge" somewhere in a contibutor's guidelines document, or just under the "Contributing" header of the README, since it is not an obvious thing to do; other projects prefer it the other way round. Or is there such a document which I failed to find?

TheZoq2:
Great work! Adding a CONTRIBUTING document stating rebasing is prefered is a pretty good idea. Though I assume that tends to almost always be the case when working on PRs, and the debate is more often about wether or not to do the final "merge" as a rebase or true merge.

Issue #288, opened 2020-12-03
impl GpioExt for GPIOF/GPIOG?


ohmyarch:
Seems to be missing.

ohmyarch:
Just found that there are no modules for gpiof/gpiog.

TheZoq2:
Good catch (and sorry for the delay), I don't think there is a good reason for us not having that already. If you're interested, I could show you how to add the impls

xieren58:
How to add the impls ? (for gpiof/gpiog...)

xieren58:
@TheZoq2

TheZoq2:
Sorry for the delay. I think all that is required is to add another one of these macros https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/gpio.rs#L1315 for the other peripherals. It probably also needs feature gating because IIRC those ports are only available on some of the chips

xieren58:
@TheZoq2 Thanks!

codytrey:
Seems that additional timers need to be implemented to fully use GPIOF/G. Since that only affects things like pwm, and adc, should this be it's own issue?

ohmyarch:
Any progress?

PR #327, opened 2021-04-10
Add GPIOF/GPIOG support

ButtNaked:
Related issue #288

Seems like high/xl density lines of different device series have pin count 80/112. 
So it's probably good enough to feature-gate like that.


ButtNaked:
Review please @TheZoq2

ohmyarch:
@TheZoq2 Please review.

burrbull:
#343

Issue #325, opened 2021-04-02
bxcan/defmt version conflict


pdgilbert:
Could you please relax the version constraint on bxcan to ">=0.4" or ">=0.4, <0.6" for reasons described at stm32-rs/stm32f7xx-hal#114 (comment).

pdgilbert:
There is also more detail at these two pull requests that have been merged.
 - https://github.com/stm32-rs/stm32f4xx-hal/pull/286
 - https://github.com/stm32-rs/stm32f7xx-hal/pull/115

PR #329, opened 2021-04-14
Update bxcan to v0.5.0

cdbrkfxrpt:
Simply updating the dependency bxcan to a version which doesn't collide with usage of defmt and friends.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ nb = "0.1.2"
 cortex-m-rt = "0.6.8"
 stm32f1 = "0.11.0"
 embedded-dma = "0.1.2"
-bxcan = "0.4.0"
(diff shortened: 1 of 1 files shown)


TheZoq2:
Awesome, thanks for taking care of this. However, as discussed in #325, it might be better to bump it to `>=0.4, <0.6` for more compatibility.

cdbrkfxrpt:
I don't disagree but 0.4 breaks usage with `defmt`. Isn't it better to just depend on the newer version in this case?

TheZoq2:
Hmm, I'm not sure, I figured there might be an opposite case where 0.5 breaks something. If nothing else, it would allow compatibility with projects which still use 0.4 I suppose.

Issue #200, opened 2020-04-13
RX DMA Access on the SPI Peripheral


Fotte:
Hi there,

I am fairly new to rust, but i have done some embedded development in the past. I could not find a forum or something similar to post my question - propably github issues are not the right way - if thats the case it would be nice if you could point me in the right direction.

I am currently trying to access an SPI ADC using DMA. I want to trigger my DMA on the reception of an external interrupt. Using the DMA i need to send and receive 5 bytes [CMD][RES0 ... RES3].

The processor side is clear to me but right now i have no clue how this could be done using this library. Could you help me out?

TheZoq2:
I believe we have only implemented SPI tx over DMA but there is no reason it couldn't be implemented for rx as well IIRC. If you would like to attempt to implement it, I'll gladly provide some pointers for where to start etc.

Fotte:
Wow, never expected such a fast answer. Yeah i would like to give it a shot in the next days. 

My main concern right now is understanding the intention behind the DMA implementation. Is there anything (Mailinglist entrys) that i could read? I'm comming from a C background so there are many rust principles that are new to me.

therealprof:
@Fotte You might want check out the Rust embedded matrix channel, https://matrix.to/#/#rust-embedded:matrix.org . There's lot's of ongoing discussion about how to implement DMA use cases properly there. Although I think a blocking implementation should be rather trivial (though potentially also not that interesting).

TheZoq2:
> Is there anything (Mailinglist entrys) that i could read?

The embedonomicon has a section on implementing DMA. https://docs.rust-embedded.org/embedonomicon/dma.html. You can also get some inspiration from the current implementations for UART, ADC and SPI TX, and the relevant PRs #104 #99, #55 and #11

Windfisch:
I worked on this in the last days and will file a PR tomorrow or so; currently, I both have implementations for receiving only, and for receiving-while-transmitting. The latter is usable, while the former does not do anything useful in SPI master mode, because without transmitting, there is no clock and without a clock, there are no bits received.
Do you think the RX-only mode is still useful (e.g. in slave mode)? I never used slave mode so far.

PR #335, opened 2021-05-16
Add DMA RX and RX/TX support to SPI

Windfisch:
Implements #200

SPI is a bit different from most other DMA-supporting peripherals, since SPI in master mode cannot receive without transmitting simultaneously. After some discussion in the #rust-embedded irc channel, we came to the conclusion that `split()`ting the device (like with serial) does not make much sense, because sending and receiving are inherently tied to each other; that made it necessary to finish the implementation of the already-existing (but unused) `RxTxDma` struct from `dma.rs` and add required traits.

This PR does that and adds an RxTxDma implementation to the SPI module.

Because in slave mode, this might(?) be still possible to receive without transmitting, I left the `RxDma` implementations in, but I am yet unsure about this.

Please do tell me if you have any suggestions :).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Support for CAN peripherals with the `bxcan` crate
 - Add DAC, UART4, UART5 clock in RCC for the f103 high density line
 - `start_raw` function and `arr`, `bsc` getters for more fine grained
-  control over the timer.
+  control over the timer
+- Added RxTxDma support support to the DMA infrastructure
+- Added DMA receive support for `SPI`
 
 ### Fixed
 - Fix > 2 byte i2c reads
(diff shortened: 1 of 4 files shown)


Windfisch:
An usage example can be found at https://github.com/Windfisch/bluepill_spidemo/blob/aa0bf2f6f24ed28cb3263e4e6f73a968538e5d70/src/main.rs#L50-L53

Windfisch:
I added a `release` function that allows to undo `with_tx_dma` etc. Also, the location of setting the actual DMA bits was changed from `SpiTxDma::start` to `with_tx_dma`.

burrbull:
What about add more type aliases?
Something like
```rust
pub type Spi1TxDma<REMAP, PINS> = SpiTxDma<SPI1, REMAP, PINS, dma::C3>;
```

burrbull:
And do we really need `SpiPayload` wrapper?

Windfisch:
nope, the SpiPayload wrapper was rather useless; I have removed it and added the type aliases you suggested

burrbull (approved):

PR #336, opened 2021-05-17
Do not enable UART DMA flags unconditionally

### LINKED ISSUE #334: UARTs should not enable DMA by default
In https://github.com/stm32-rs/stm32f1xx-hal/blob/5d55f83528a1a4eebcc672201751a320d9be5854/src/serial.rs#L312, every U(S)ART enables the DMA transmit and receive bits (dmat and dmar) unconditionally when initializing the UART struct.

This interferes with other users of the DMA channels, and it appears to me that this should rather be done in `TransferPayload::start()/::stop()`: https://github.com/stm32-rs/stm32f1xx-hal/blob/5d55f83528a1a4eebcc672201751a320d9be5854/src/serial.rs#L530-L546), like in the SPI implementation https://github.com/stm32-rs/stm32f1xx-hal/blob/5d55f83528a1a4eebcc672201751a320d9be5854/src/spi.rs#L569-L586

Or did I overlook something here?

Windfisch:
Fixes #334

I have not tested that code, though, so proceed with caution :)

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -563,7 +563,11 @@ macro_rules! serialdma {
             }
 
             impl $rxdma {
+                #[deprecated(since = "0.7.1", note = "Please use release instead")]
                 pub fn split(mut self) -> (Rx<$USARTX>, $dmarxch) {
+                    self.release()
+                }
+                pub fn release(mut self) -> (Rx<$USARTX>, $dmarxch) {
                     self.stop();
                     unsafe { (*$USARTX::ptr()).cr3.write(|w| w.dmar().clear_bit()); }
                     let RxDma {payload, channel} = self;
@@ -575,7 +579,11 @@ macro_rules! serialdma {
             }
 
             impl $txdma {
+                #[deprecated(since = "0.7.1", note = "Please use release instead")]
                 pub fn split(mut self) -> (Tx<$USARTX>, $dmatxch) {
+                    self.release()
+                }
+                pub fn release(mut self) -> (Tx<$USARTX>, $dmatxch) {
                     self.stop();
                     unsafe { (*$USARTX::ptr()).cr3.write(|w| w.dmat().clear_bit()); }
                     let TxDma {payload, channel} = self;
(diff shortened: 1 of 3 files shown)


Windfisch:
Updated as per https://github.com/stm32-rs/stm32f1xx-hal/issues/334#issuecomment-842916316.

Windfisch:
Renamed serial-dma `split` and i2c `free` functions to `release` as per https://github.com/stm32-rs/stm32f1xx-hal/issues/334#issuecomment-843724427.

src/serial.rs
```diff
@@ -564,8 +563,13 @@ macro_rules! serialdma {
             }
 
             impl $rxdma {
+                #[deprecated(since = "0.7.1", note = "Please use release instead")]
                 pub fn split(mut self) -> (Rx<$USARTX>, $dmarxch) {
```
TheZoq2, line 567: Could you remove the `mut` here since it gives a warning now that self is just passed along?

src/serial.rs
```diff
@@ -575,8 +579,13 @@ macro_rules! serialdma {
             }
 
             impl $txdma {
+                #[deprecated(since = "0.7.1", note = "Please use release instead")]
                 pub fn split(mut self) -> (Tx<$USARTX>, $dmatxch) {
```
TheZoq2, line 583: Remove mut here as well

TheZoq2 (changes requested):
Thanks for the PR! I pointed out a few `mut`s which cause warnings, other than that I think this is ready to merge (possibly after testing on hw)

burrbull:
cc @TeXitoi

TeXitoi (approved):

burrbull:
> @burrbull you want acces right to this repo?

It would be great.

TeXitoi:
I can't find how to add you to the stm32f1 team, looks like I don't have the right or I can't find the settings

adamgreig:
Previously only @therealprof was a maintainer for the stm32f1 team, but I've added you as well @TeXitoi, so you should be able to add new team members now.

TeXitoi:
Thanks, @burrbull added.

burrbull:
Thanks.

Issue #339, opened 2021-06-04
impl spi ?


pdgilbert:
(Possibly related to #118)
I am trying to use a `setup()` function to return a `spi` object and the `cs` pin to use in my main code. I can do this with specific types for the return object like this stripped down (but building ok) example:
<details>
<summary>Click to expand</summary>

```
#![deny(unsafe_code)]
#![no_std]
#![no_main]

use cortex_m_rt::entry;
use panic_rtt_target as _;

use embedded_hal::spi::{Mode, Phase, Polarity};
pub const MODE: Mode = Mode {
    phase: Phase::CaptureOnSecondTransition,
    polarity: Polarity::IdleHigh,
};

#[cfg(feature = "stm32f1xx")]
use stm32f1xx_hal::{
    gpio::{
        gpioa::{PA4, PA5, PA6, PA7},
        Alternate, Floating, Input,
    },
    gpio::{Output, PushPull},
    pac::{Peripherals, SPI1},
    prelude::*,
    spi::{Spi, Spi1NoRemap, Pins},
};

#[cfg(feature = "stm32f1xx")]
fn setup() -> (Spi<SPI1, Spi1NoRemap, (PA5<Alternate<PushPull>>, PA6<Input<Floating>>, PA7<Alternate<PushPull>>), u8>,
    PA4<Output<PushPull>>) {
//fn setup() -> (Spi<SPI1, Spi1NoRemap, impl Pins<Spi1NoRemap, SPI1>, u8>, PA4<Output<PushPull>>) {

    let dp = Peripherals::take().unwrap();

    let mut flash = dp.FLASH.constrain();
    let mut rcc = dp.RCC.constrain();

    let clocks = rcc.cfgr.freeze(&mut flash.acr);

    let mut afio = dp.AFIO.constrain(&mut rcc.apb2);
    let mut gpioa = dp.GPIOA.split(&mut rcc.apb2);

    // SPI1
    let  sck = gpioa.pa5.into_alternate_push_pull(&mut gpioa.crl);
    let miso = gpioa.pa6;
    let mosi = gpioa.pa7.into_alternate_push_pull(&mut gpioa.crl);
    let  cs  = gpioa.pa4.into_push_pull_output(&mut gpioa.crl);

    let spi = Spi::spi1(
        dp.SPI1,
        (sck, miso, mosi),
        &mut afio.mapr,
        MODE,
        1_u32.mhz(),
        clocks,
        &mut rcc.apb2,
[...]

PR #340, opened 2021-06-06
SCK, MISO, MOSI

burrbull:
Use only this sequence.

Related to #339.
Need decision. What we need more? Possibility to pass pins in custom sequence or more simple API.
cc @TheZoq2 @therealprof @TeXitoi @adamgreig @thalesfragoso

### DIFF
--- a/examples/spi.rs
+++ b/examples/spi.rs
@@ -1,0 +1,60 @@
+#![deny(unsafe_code)]
+#![no_std]
+#![no_main]
+
+use cortex_m_rt::entry;
+use panic_halt as _;
+
+use embedded_hal::spi::{Mode, Phase, Polarity};
+pub const MODE: Mode = Mode {
+    phase: Phase::CaptureOnSecondTransition,
+    polarity: Polarity::IdleHigh,
+};
+
+use stm32f1xx_hal::{
+    gpio::gpioa::PA4,
+    gpio::{Output, PushPull},
+    pac::{Peripherals, SPI1},
+    prelude::*,
+    spi::{Pins, Spi, Spi1NoRemap},
+};
+
+fn setup() -> (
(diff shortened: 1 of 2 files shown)


burrbull:
Also we could add advanced trait
```rust
pub trait SpiExt<SPI> {}

impl<SPI, REMAP, PINS> SpiExt<SPI> for Spi<SPI1, REMAP, PINS, u8>
where
    REMAP: Remap<Periph = SPI>,
    PINS: Pins<REMAP>,
{
}
```
that would let write just
```rust
fn setup() -> (impl SpiExt<SPI1>, PA4<Output<PushPull>>) {
```
But this is rare needed case.

TeXitoi (approved):

TeXitoi:
I'm fine with removing the free order agrument.

pdgilbert:
FWIW, the PR is working (using `Spi<SPI1, Spi1NoRemap, impl Pins<Spi1NoRemap>, u8>,`) in my testing too.

burrbull:
bors r=TeXitoi

TeXitoi:
I think there is no bors here.

burrbull:
> I think there is no bors here.

Do you have permissions for manual merge?

TeXitoi:
Yes, that's a long time I didn't to it here. rebase and merge?

TeXitoi:
Looks more like squash and merge here.

burrbull:
> Yes, that's a long time I didn't to it here. rebase and merge?

Just merge.

Could we add bors?

TeXitoi:
I have nothing against that. @TheZoq2 ?

TheZoq2:
Looks like I'm late, but I have nothing against that. I'll have to learn how to use it, but that should be no problem :)

Re: merge/rebase/squash, I like to keep a linear history so I tend to squash PRs with small changes and many random commits, and rebase the rest.

PR #343, opened 2021-07-09
Const generic GPIOs

burrbull:


therealprof:
@TheZoq2 This is similar to the change landed in the `stm32f4xx-hal` and to me it marks a huge improvement. What's your opinion?

src/gpio.rs
```diff
@@ -175,6 +202,89 @@ pub trait ExtiPin {
...
+                    .modify(|r, w| unsafe { w.bits(r.bits() | (1 << i)) });
+                exti.rtsr
+                    .modify(|r, w| unsafe { w.bits(r.bits() & !(1 << i)) });
+            }
+            Edge::RISING_FALLING => {
+                exti.rtsr
+                    .modify(|r, w| unsafe { w.bits(r.bits() | (1 << i)) });
+                exti.ftsr
+                    .modify(|r, w| unsafe { w.bits(r.bits() | (1 << i)) });
+            }
```
TheZoq2, line 261: Looks like we now use a runtime value determine the constant to write, is that constant folded away? I.e. is the compiler smart enough to figure out that `self.pin_id` is constant?

src/gpio.rs
```diff
@@ -293,1040 +377,841 @@ macro_rules! gpio {
...
-                fn is_set_high(&self) -> Result<bool, Self::Error> {
-                    self.is_set_low().map(|b| !b)
-                }
-
-                fn is_set_low(&self) -> Result<bool, Self::Error> {
-                    // NOTE(unsafe) atomic read with no side effects
-                    Ok(unsafe { (*$GPIOX::ptr()).odr.read().bits() & (1 << self.i) == 0 })
+            impl<MODE> PEPin<MODE, $port_id> {
+                pub fn erase(self) -> EPin<MODE> {
+                    EPin::$PXx(self)
```
TheZoq2, line 391: I don't really like calling this EPin, to me, it's not obvious that that stands for erased pin, should we just call it `ErasedPin` instead?

src/gpio.rs
```diff
@@ -293,1040 +377,841 @@ macro_rules! gpio {
...
+        }
+    }
+}
 
-            pub type $PXx<MODE> = Pxx<MODE>;
-
-
-            impl<MODE> Mode<MODE> for Generic<MODE> {}
+/// Pin
+pub struct Pin<MODE, CR, const P: char, const N: u8> {
```
TheZoq2, line 412: Another naming question: Should we stick with P and N here, or just write out `Port` and `Number`. The latter is clearly more verbose, but also makes it less ambiguous

src/gpio.rs
```diff
@@ -293,1040 +377,841 @@ macro_rules! gpio {
...
-
-                        $PXi { mode: Alternate::_new() }
-                    }
+    #[inline(always)]
+    fn pin_id(&self) -> u8 {
+        N
+    }
+    #[inline(always)]
+    fn port_id(&self) -> u8 {
+        P as u8 - 0x41
```
TheZoq2, line 437: Where does this 0x41 come from?

src/gpio.rs
```diff
@@ -293,1040 +377,841 @@ macro_rules! gpio {
...
-                    #[inline]
-                    pub fn into_open_drain_output(
-                        self,
-                        cr: &mut $CR,
-                    ) -> $PXi<Output<OpenDrain>> {
-                        self.into_open_drain_output_with_state(cr, State::Low)
-                    }
+    #[inline(always)]
+    fn _is_set_low(&self) -> bool {
+        unsafe { (*Gpio::<P>::ptr()).odr.read().bits() & (1 << N) == 0 }
```
TheZoq2, line 517: Missing `//NOTE(unsafe)` comment

src/gpio.rs
```diff
@@ -293,1040 +377,841 @@ macro_rules! gpio {
...
-                        mut self,
-                        cr: &mut $CR,
-                        initial_state: State,
-                    ) -> $PXi<Output<PushPull>> {
-                        self.set_state(initial_state);
-                        unsafe {
-                            $PXi::<Output<PushPull>>::set_mode(cr)
-                        }
-                    }
+// embedded_hal impls
```
TheZoq2, line 537: This comment looks wrong, we are not `impl`ing `embedded_hal` in this block, right?

src/gpio.rs
```diff
@@ -293,1040 +377,841 @@ macro_rules! gpio {
...
+impl<MODE, CR, const P: char, const N: u8> Pin<Output<MODE>, CR, P, N> {
```
TheZoq2, line 538: I haven't tested this, but what happens if you `use embedded_hal::digital::v2::OutputPin`, will the compiler not throw an "ambiguous function call error" since we have both the local `set_high` and the OutputPin set_high in scope.

Is this a problem? Not having to `unwrap` certainly has its advantages, but at the same time, it would make code like

```rust

fn generic_fn(pin: impl OutputPin) {
    // ...
}

fn main() {
    let pin = // get a pin
    pin.set_high(); // Now ambiguous?
}
```

break, which might be confusing if the generic stuff is added later in a project.

Ok, I tested it in the playground and it seems like this is a non-issue. I guess I'm not entirely sure how method resolution works in rust.. I'll leave this comment here in case it is an issue in this case, but feel free to ignore it.

src/gpio.rs
```diff
@@ -293,1040 +377,841 @@ macro_rules! gpio {
...
 
-                        cr.cr().modify(|r, w| unsafe {
-                            w.bits((r.bits() & !(0b11 << OFFSET)) | ((speed as u32) << OFFSET))
-                        });
-                    }
-                }
+        impl<const P: char> Cr<$CR, P> {
+            // NOTE(allow) we get a warning on GPIOC because it only has 3 high pins
+            #[allow(dead_code)]
+            pub(crate) fn cr(&mut self) -> &pac::gpioa::$CR {
```
TheZoq2, line 678: Maybe I'm misunderstanding things, but should this really be `gpioa`, seems like this is more generic?

src/gpio.rs
```diff
@@ -293,1040 +377,841 @@ macro_rules! gpio {
...
+                const MODE: u32 = 0b11;
+                const BITS: u32 = (CNF << 2) | MODE;
+
+                // input mode
+                cr.cr().modify(|r, w| unsafe {
+                    w.bits((r.bits() & !(0b1111 << Self::OFFSET)) | (BITS << Self::OFFSET))
+                });
+
+                Pin::new(Alternate::_new())
+            }
```
TheZoq2, line 727: Why can these not use the `set_mode` functions?

TheZoq2 (commented):
Sorry for the delay, as you can probably tell I've been distracted by other things.

Looks like a nice change overall, I left some review comments for the main gpio.rs file. I don't really have the energy to review the erased and partially erased files right now since it's 10:30 pm, I'll see if I can get that done tomorrow

src/gpio.rs
burrbull: Yes, compiler optimize out this match to same code as before in release mode.

src/gpio.rs
burrbull: This is `A` to 0 conversion.

src/gpio.rs
burrbull: Ok

src/gpio.rs
burrbull: It will be too long to use it everywhere. I will add it to docs better.

src/gpio.rs
burrbull: Sorry. Didn't notice.

src/gpio.rs
burrbull: Ok

src/gpio.rs
burrbull: Maybe

burrbull:
Talking about erased pin we have alternative implementation for it:
https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/gpio/erased.rs
It is not same beautiful but should be more optimized.

burrbull:
@TheZoq2 Can we merge this now?

TheZoq2 (commented):
I don't think we should add external crates for dependencies, apart from that I think this looks good

Cargo.toml
```diff
@@ -52,6 +52,7 @@ mfrc522 = "0.2.0"
 serde_derive = "1.0.90"
 usb-device = "0.2.3"
 usbd-serial = "0.1.0"
+hd44780-driver = "0.4.0"
```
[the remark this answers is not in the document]
burrbull: Ok. I've removed last commit.

talhaHavadar:
Is there an alternative for downgrade function?

PR #344, opened 2021-07-13
Rcc bit-band

burrbull:
~~Depends on #337~~

### DIFF
--- a/src/afio.rs
+++ b/src/afio.rs
@@ -1,7 +1,7 @@
 //! # Alternate Function I/Os
-use crate::pac::{afio, AFIO};
+use crate::pac::{afio, AFIO, RCC};
 
-use crate::rcc::{Enable, Reset, APB2};
+use crate::rcc::{Enable, Reset};
 
 use crate::gpio::{
     gpioa::PA15,
@@ -10,13 +10,14 @@ use crate::gpio::{
 };
 
 pub trait AfioExt {
-    fn constrain(self, apb2: &mut APB2) -> Parts;
+    fn constrain(self) -> Parts;
 }
 
 impl AfioExt for AFIO {
-    fn constrain(self, apb2: &mut APB2) -> Parts {
-        AFIO::enable(apb2);
-        AFIO::reset(apb2);
+    fn constrain(self) -> Parts {
+        let rcc = unsafe { &(*RCC::ptr()) };
+        AFIO::enable(rcc);
+        AFIO::reset(rcc);
 
         Parts {
             evcr: EVCR { _0: () },
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -10,6 +10,8 @@ use crate::time::Hertz;
 
 use crate::backup_domain::BackupDomain;
 
+mod enable;
+
 /// Extension trait that constrains the `RCC` peripheral
 pub trait RccExt {
     /// Constrains the `RCC` peripheral so it plays nicely with the other abstractions
@@ -19,9 +21,6 @@ pub trait RccExt {
 impl RccExt for RCC {
     fn constrain(self) -> Rcc {
         Rcc {
-            ahb: AHB { _0: () },
-            apb1: APB1 { _0: () },
-            apb2: APB2 { _0: () },
             cfgr: CFGR {
                 hse: None,
                 hclk: None,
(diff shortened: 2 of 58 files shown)


TheZoq2:
This seems like a super ergonomic change! I haven't gone through the code in detail, but from an overview, it looks OK from a safety perspective (assuming the bit banding thing is atomic, which it is as far as I can tell)

burrbull:
Rebased.

burrbull:
Removed dependency from #337

burrbull:
cc @TheZoq2 I think it's ready to merge.

src/afio.rs
```diff
@@ -10,13 +10,16 @@ use crate::gpio::{
...
 impl AfioExt for AFIO {
-    fn constrain(self, apb2: &mut APB2) -> Parts {
-        AFIO::enable(apb2);
-        AFIO::reset(apb2);
+    fn constrain(self) -> Parts {
+        unsafe {
+            let rcc = &(*RCC::ptr());
+            AFIO::enable(rcc);
+            AFIO::reset(rcc);
+        }
```
TheZoq2, line 22: As far as I can tell, without doing this, any use of AFIO will be pointless, but in some of the examples, we no longer constrain AFIO.

Note to self: Is this the case?

src/rcc.rs
```diff
@@ -45,92 +44,58 @@ impl RccExt for RCC {
...
 
 impl AHB {
-    // TODO remove `allow`
-    #[allow(dead_code)]
-    pub(crate) fn enr(&mut self) -> &rcc::AHBENR {
-        // NOTE(unsafe) this proxy grants exclusive access to this register
-        unsafe { &(*RCC::ptr()).ahbenr }
+    fn enr(rcc: &RccRB) -> &rcc::AHBENR {
+        &rcc.ahbenr
     }
```
TheZoq2, line 59: Are these functions required anmyore, would it not be neater to just do `rcc.ahbenr` directly when we do need them?

src/rcc.rs
```diff
@@ -576,191 +543,26 @@ impl GetBusFreq for APB2 {
...
 }
 
+use crate::pac::rcc::RegisterBlock as RccRB;
```
TheZoq2, line 546: I don't like this alias, it doesn't seem like just typing out `RegisterBlock` directly in a few places is that annoying, and saves some jumping around when reading the code (unless there are multiple `RegisterBlock`s in scope here, though in that case I'd prefer to just write out `rcc::RegisterBlock`

src/rcc/enable.rs
```diff
@@ -0,0 +1,185 @@
+use super::*;
+use crate::bb;
+
+macro_rules! bus {
+    ($($PER:ident => ($apbX:ty, $bit:literal),)+) => {
+        $(
+			impl Sealed for crate::pac::$PER {}
```
TheZoq2, line 7: Weird spacing (is this a rustfmt oddity?)

src/rcc/enable.rs
```diff
@@ -0,0 +1,185 @@
...
+bus! {
+    TIM9 => (APB2, 19),
+    TIM10 => (APB2, 20),
+    TIM11 => (APB2, 21),
+}
+
+#[cfg(any(feature = "stm32f102", feature = "stm32f103"))]
+bus! {
+    USB => (APB1, 23),
+}
```
TheZoq2, line 185: Note to self, check these mappings

src/timer.rs
```diff
@@ -281,10 +280,13 @@ macro_rules! hal {
...
                 /// Initialize timer
-                pub fn $timX(tim: $TIMX, clocks: &Clocks, apb: &mut $APBx) -> Self {
+                pub fn $timX(tim: $TIMX, clocks: &Clocks) -> Self {
                     // enable and reset peripheral to a clean slate state
-                    $TIMX::enable(apb);
-                    $TIMX::reset(apb);
+                    unsafe {
+                        let rcc = &(*RCC::ptr());
+                        $TIMX::enable(rcc);
+                        $TIMX::reset(rcc);
```
TheZoq2, line 288: This note applies to all copies of `enable` and `reset` function calls: Should we not move them out of the unsafe block? They are safe to call so they don't need to be there as far as I can tell?

On the other hand, does it make sense to not have them be unsafe. We could make them public API and mark them as unsafe in case people want to circumvent part of the HAL in a project but still want to init the peripherals via RCC

TheZoq2 (commented):
Looks very good overall, I left a few notes with minor things to change, and a few things for me to look into after lunch.

src/afio.rs
burrbull: Not sure what you mean, but `&mut afio.mapr` stayed where it was. I've not deleted it.

src/rcc.rs
burrbull: These functions are used in `bus!` macro.

src/afio.rs
TheZoq2: I was (wrongly) under the impression that AFIO was used for some GPIO stuff, but that seems to not be the case. Disregard this comment

src/rcc.rs
TheZoq2: Ah right, good point

src/rcc.rs
burrbull: Fixed.

src/rcc/enable.rs
burrbull: Fixed.

src/timer.rs
burrbull: > This note applies to all copies of `enable` and `reset` function calls: Should we not move them out of the unsafe block? They are safe to call so they don't need to be there as far as I can tell?

Moved.

burrbull: > On the other hand, does it make sense to not have them be unsafe. We could make them public API and mark them as unsafe in case people want to circumvent part of the HAL in a project but still want to init the peripherals via RCC

We can, but I don't see sense to make them unsafe as `rcc` can be gotten in safe way only during initialization. In all other places we need to use unsafe `RCC::ptr()`.

TheZoq2:
One small final note, these changes should go into Breaking Changes in the changelog, other than that this looks ready for squash and merge

burrbull:
Done.

PR #346, opened 2021-07-14
clippy

burrbull:
Thanks for contributing to the project!

Please write a short description of your changes.

We also maintain a changelog in CHANGELOG.md. If you add a feature or fix a bug, please add an entry there :)


TheZoq2:
Should we consider adding clippy to CI?

PR #319, opened 2021-03-11
Add 9-bit word length mode for serial

AndreySmirnov81:
Closes #262

PR #355, opened 2021-08-04
Add 9-bits word length mode for serial (new)

### LINKED ISSUE #262: Support for 9 databits for USART peripheral
Some STM32F1 devices (at least stm32f103) supports programmable data word lengths (8 or 9 bits) for USART peripheral:
```
Word length may be selected as being either 8 or 9 bits by programming the M bit in the
USART_CR1 register.
```

image

burrbull:
~~Depends on #354~~

Instead of #319 
Closes #262 

cc @AndreySmirnov81

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


AndreySmirnov81:
@burrbull maybe also close the branch [serial_9b2](https://github.com/stm32-rs/stm32f1xx-hal/tree/serial_9b2)

PR #360, opened 2021-08-10
Add serial is_rxne is_idle is_txe 

gqf2008:
Add serial idle method and example


burrbull:
Could you rename them to more verbose `is_tx_empty` and `is_rx_not_empty`?

PR #362, opened 2021-08-16
Fix failing flash verification on erase

Windfisch:
When erasing a 2kb flash page, flash erase verification wanted all `u16`s with addresses starting from 0 to 2047 to be `0xFFFF`.
This includes the u16 located from 2047 to 2048 (inclusive), which contains the last byte of the to-be-erased page and the first byte of the next page.
Surely, this u16 cannot reasonably be expected to be 0xFFFF after the erase.

Stepping over even addresses only solves this issue.


burrbull (approved):

burrbull:
Same as #328 

bors r+

Windfisch:
That was quick. Thanks <3

PR #363, opened 2021-08-17
add uart4 uart5

apeng2012:
USART123 use crate::pac::usart1::RegisterBlock
UART45 use crate::pac::uart4::RegisterBlock

Cannot use the same Instance [ffba652](https://github.com/stm32-rs/stm32f1xx-hal/commit/ffba652ca0311a8e216386638adc29a9be5e536b)

I use the previous macro to increase UART45


burrbull:
Close in favor of #529

PR #365, opened 2021-10-11
Use max frequency and round down when calculating PCLK1

jasLogic:
When pclk1 is not set use hclk or 36 MHz to stay below the max frequency of PCLK1.
This is needed if you set the hclk to a value > 36 MHz and don't specify a value for pclk1. Then the prescaler would be set to one automatically and the maximum value would be exceeded. 

Also round up when calculating the prescaler so that the given pclk1 value is not exceeded.
This is important because you could have configurations where the prescaler would still be too small to stay below 36 MHz. For example when hclk = 48 MHz and pclk1 = 36 Mhz than hclk / pclk1 = 1 (integer division is rounded down) and the prescaler would be set to one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Renamed `serial`'s `RxDma`/`TxDma`'s `split` method into `release`
 - Renamed I2C's `free` method into `release`
 - Enable SPI DMA in `with_tx_dma`, not in `SpiTxDma::start`
+- Use maximum frequency of 36 MHz on PCLK1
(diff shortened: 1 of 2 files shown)


TheZoq2:
Thanks for the PR, seems like a great change that should prevent some weird behaviour!

Since #337 is still open, I think there will be a conflict with it, @burrbull does it make sense to move this change over there and merge the whole thing in one go? (I guess  I also still need to have a look at that PR :D)

jasLogic:
Btw: Should I also add this to the changelog? The commit doesn't change the API but I just thought that it might round specified frequency down instead of up (like it did before) which could break a user program.

PR #372, opened 2021-11-14
Fix, enable PWR clock before accessing the PWR_CR

apeng2012:
dbp can't be written to unless PWRen has been set.


burrbull:
Add to changelog, please

burrbull (approved):

PR #373, opened 2021-11-16
Use modify instead of write

gnxlxnxx:
If we use write here one can not use dma with tx and rx at the same time, as the second with_dma() overwrites the register


burrbull (approved):

burrbull:
Do same in `release` methods.

gnxlxnxx:
ah, missed that

burrbull (approved):

TheZoq2:
LGTM, Thanks!

PR #377, opened 2021-12-08
RTC clock source can be selected

apeng2012:
It has been verified that all three RTC clock sources can work.

### DIFF
--- a/src/rtc.rs
+++ b/src/rtc.rs
@@ -7,9 +7,18 @@ use crate::backup_domain::BackupDomain;
 use crate::time::Hertz;
 
 use core::convert::Infallible;
+use core::marker::PhantomData;
 
 // The LSE runs at at 32 768 hertz unless an external clock is provided
 const LSE_HERTZ: u32 = 32_768;
+const LSI_HERTZ: u32 = 40_000;
+
+/// RTC clock source HSE clock divided by 128 (type state)
+pub struct RtcClkHseDiv128;
+/// RTC clock source LSE oscillator clock (type state)
+pub struct RtcClkLse;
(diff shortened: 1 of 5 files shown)


src/rtc.rs
```diff
@@ -28,11 +37,12 @@ const LSE_HERTZ: u32 = 32_768;
   [examples/blinky_rtc.rs]: https://github.com/stm32-rs/stm32f1xx-hal/blob/v0.7.0/examples/blinky_rtc.rs
 */
 
-pub struct Rtc {
+pub struct Rtc<CS> {
```
burrbull, line 40: Default type could be added
```suggestion
pub struct Rtc<CS=RtcClkLse> {
```

src/rtc.rs
```diff
@@ -44,9 +54,12 @@ impl Rtc {
       reset the counter.
     */
     pub fn rtc(regs: RTC, bkp: &mut BackupDomain) -> Self {
```
burrbull, line 56:
```suggestion
    pub fn new(regs: RTC, bkp: &mut BackupDomain) -> Self {
```

src/rtc.rs
```diff
@@ -44,9 +54,12 @@ impl Rtc {
...
-        let mut result = Rtc { regs };
+        let mut result = Rtc {
+            regs,
+            _clock_source: PhantomData,
+        };
 
-        Rtc::enable_rtc(bkp);
+        Rtc::<RtcClkLse>::enable_rtc(bkp);
```
burrbull, line 62:
```suggestion
        Self::enable_rtc(bkp);
```

burrbull:
Looks great in general

burrbull (approved):

PR #385, opened 2022-01-07
Add support for reconfiguring a Serial port

horazont:
Example use case is to implement a [1-Wire Bus "Master"][1]. In that example, the UART is used instead of bit-banging the 1-Wire protocol. However, not all signals can be generated with a single baud rate. In particular, the 1-wire reset strobe is much longer than the 1-wire data strobes, requiring to use a different, slower baud rate to reset devices.

   [1]: https://www.maximintegrated.com/en/design/technical-documents/tutorials/2/214.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- Serial can now be reconfigured, allowing to change e.g. the baud rate after initialisation.
+
 ## [v0.8.0] - 2021-12-29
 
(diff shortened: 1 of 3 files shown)


burrbull:
LGTM.
rustfmt, please.
And example maybe?

horazont:
@burrbull Thanks for the feedback.

I fmt'd the code and added a simple example. A full UART-based 1-wire master would be quite complex and would fit more into a separate library.

Tested it with the STM32VLDISCOVERY board (had to adapt memory.x and use breakpoints instead of asm::bkpt locally).

burrbull (approved):

PR #388, opened 2022-01-08
timer Instance, rtic monotonic

burrbull:
Source:
https://github.com/jhillyerd/hw-cpu/blob/main/firmware/src/mono.rs


therealprof (approved):
LGTM, thanks.

bors r+

PR #390, opened 2022-01-09
Implement Rx/Tx reunification for Serial

### LINKED ISSUE #386: Add reunite for Serial Tx/Rx
With some functions (e.g. reconfiguration) only safely available when the peripherial is fully under control of the caller, it would make sense to be able to recombine the Tx and Rx parts after a split.

`reunite` [1] was attempted in #36 (under the name "release"), however, it was closed because the ReleaseToken was deemed unnecessary. Upon attempting an implementation without the ReleaseToken, I came into a questionable situation:

To avoid the release token: 

- the `PINS` type information must be carried on to the `Tx` and `Rx` instances somehow.
- the `USART` and `PINS` instances need to be carried on to either `Tx` or `Rx` in order to be able to recover them when reuniting

These points can be addressed by moving the `PINS` and `USART` instances to `Tx` and adding a PINS PhantomData to the `Rx` (or vice versa).

However: This has fallout to the `{T,R}x{Dma,}{1,2,3}` type aliases, which then become generic over PINS. That seemed somehow excessive and is for sure a breaking change for downstream code, so I wanted to run this by you folks before venturing further down this direction.

- Should the type aliases be generalized or should there be explicit aliases for the two options for each USART?
- Should we move ahead with the release-token-free way or does this caveat warrant keeping that token (which I personally also find a bit awkward)?
- If not generalizing the type aliases, what would be good names? `Tx1` and `Tx1Alt`?

   [1]: The name `reunite` has prior art [in the owned halves of of `tokio::net::TcpStream`](https://docs.rs/tokio/latest/tokio/net/tcp/struct.OwnedReadHalf.html#method.reunite). There is also [`unsplit` in tokio](https://docs.rs/tokio/latest/tokio/io/struct.ReadHalf.html#method.unsplit), but given the choice I'd prefer `reunite`.

horazont:
There are many possible pin combinations for a Serial peripheral. Many support at least one alternative pin group to be used, and in addition to that, each pin can also be configured differently. For instance, instead of the common `Output<PushPull>` with an `Input<Floating>`, one could also use `Output<OpenDrain>` and `Input<PullUp>`, e.g. when
interfacing with a 1-Wire bus.

This pin information is already lost when splitting the Serial object into `Rx`/`Tx`. In order to allow a reunification into a Serial object which represents ownership of both `Tx` and `Rx`, we need a Serial variant which has that pin information erased. 

This PR introduces `ErasedSerial`, which serves exactly that purpose. In addition, `Tx` and `Rx` both gain a consuming `reunite` function which restores the original `Serial`. To facilitate that and provide the original contents of the `ErasedSerial` struct, we use the `Tx` to stash away the USART instance -- this choice is arbitrary. As the USART is zero-sized anyway and we get both parts (Rx and Tx) when reuniting, it shouldn't matter where we stash it away.

Fixes #386.

### DIFF
--- a/examples/serial.rs
+++ b/examples/serial.rs
@@ -88,5 +88,13 @@ fn main() -> ! {
     assert_eq!(received, sent);
     asm::bkpt();
 
+    // And later reunite it again
+    let mut serial = tx.reunite(rx);
+    let sent = b'Z';
+    block!(serial.write(sent)).ok();
+    let received = block!(serial.read()).unwrap();
+    assert_eq!(received, sent);
+    asm::bkpt();
+
     loop {}
 }
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -185,8 +185,12 @@ use crate::pac::usart1 as uart_base;
 
 /// Serial abstraction
 pub struct Serial<USART, PINS> {
-    usart: USART,
     pins: PINS,
+    inner: ErasedSerial<USART>,
+}
+
+pub struct ErasedSerial<USART> {
+    usart: USART,
     tx: Tx<USART>,
     rx: Rx<USART>,
 }
@@ -224,7 +228,7 @@ impl<USART> Tx<USART> {
     }
 }
 
-impl<USART, PINS> Serial<USART, PINS>
+impl<USART> ErasedSerial<USART>
 where
     USART: Instance,
 {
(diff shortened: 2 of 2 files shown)


horazont:
Note that this answers some of the design questions in #386 in code. I am using this successfully in my async/await experiments on the STM32VLDISCOVERY board, but I'm open for changing things if you disagree about anything.

burrbull:
Possibly if Tx contain tx_pin and Rx contain rx_pin we will not need ErasedSerial

burrbull:
See #392

burrbull:
cc @TheZoq2

burrbull:
Is reconfigure the main goal of reunification for you?

burrbull:
> Is reconfigure the main goal of reunification for you?

In this case I can propose easier solution: #393 
The advantage of external function that it takes mutable references instead of owning Tx and Rx and can be used in RTIC, etc.

PR #391, opened 2022-01-09
replace GetBusFreq with BusClock

burrbull:
Same as in f3xx-hal, f4xx-hal


therealprof (approved):
LGTM, thanks

bors r+

PR #392, opened 2022-01-09
split Serial PINS

burrbull:
Alternative to #390

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Possibility to unite Tx and Rx parts of Serial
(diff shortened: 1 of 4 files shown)


horazont (commented):
Right, that would work. However, I think this points at a problem with the API: https://github.com/stm32-rs/stm32f1xx-hal/pull/392/files#diff-6c6f014a7f636e0dd1ce76320128a53b7bed96a29540822c340aeba0d9c9c1fdR23-R24

Now, the interrupt implementation (which cannot be easily generalized because it needs a `static`) depends on the pins used for the serial peripheral. Given that, I'd still want to introduce a pin-erased version of Serial, Tx and Rx. Nonetheless, I agree that #392 is a better standalone fix for #386 than #390.

PR #400, opened 2022-02-14
serial refactoring

AndreySmirnov81:
@burrbull

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -76,12 +76,10 @@
 //! let received: u8 = block!(rx.read()).unwrap();
 //!  ```
 
+use core::convert::Infallible;
 use core::marker::PhantomData;
 use core::ops::Deref;
 use core::sync::atomic::{self, Ordering};
-
-use crate::pac::{RCC, USART1, USART2, USART3};
-use core::convert::Infallible;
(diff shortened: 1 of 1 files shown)


src/serial.rs
```diff
@@ -320,21 +319,19 @@ where
...
-                .pce()
-                .bit(parity_control_enable)
+        self.usart.cr1.modify(|_r, w| { w
+            .m().bit(match config.wordlength {
+                WordLength::DataBits8 => false,
+                WordLength::DataBits9 => true,
+            })
+            .ps().bit(parity_is_odd)
+            .pce().bit(parity_is_used)
         });
```
burrbull, line 335: If you want to write it in 1 line you can use this way
```suggestion
        self.usart.cr1.modify(|_r, w| {
            w.m().bit(match config.wordlength {
                WordLength::DataBits8 => false,
                WordLength::DataBits9 => true,
            });
            w.ps().bit(parity_is_odd);
            w.pce().bit(parity_is_used)
        });
```
or rustfmt

burrbull:
LGTM. Please squash commits

src/serial.rs
AndreySmirnov81: Thanks for the hint

PR #401, opened 2022-02-16
use fugit::Rate types instead of custom

burrbull:
Goodby, `Into<Hertz>`

Remove `cast`


therealprof (approved):
Let's do this! 🎉

bors r+

PR #402, opened 2022-02-18
Rename Serial items

AndreySmirnov81:
Maybe do the following renaming?

wordlength_8 -> wordlength_8bits
wordlength_9 -> wordlength_9bits

(since there may be not only data bits but also parity)
WordLength::DataBits8 -> WordLength::Bits8
WordLength::DataBits9 -> WordLength::Bits9

And also add 
Config::wordlength(mut self, wordlength: WordLength).


burrbull:
Don't have strict opinion on naming. So if you think it is better I'm not opposite.

burrbull (approved):

PR #404, opened 2022-02-19
timers rework

burrbull:


aboarya:
@burrbull I updated the `Cargo.tml` to point to this branch.  The code compiles and flashes the chip with no problems, however while observing the LED I am noticing different behavior than what's programmed.  For instance

```rust
loop {
            // On for 10s, off for 30s.
            led.set_high();
            // Use `embedded_hal::DelayMs` trait
            delay.delay_ms(10000_u32);
            led.set_low();
            // or use `fugit` duration units
            delay.delay(30.secs());
        }
```
If I understood the comments correctly the LED should remain on for `10s` and off for `30s`, however I am observing the opposite.  The LED is on for `30s` and off for `10s`. 

Is this just user error on my part?

burrbull:
It is just how your LED is connected on scheme.
`set_high` mean set +3.3 V on pin, `set_low` mean set 0 V.
If your LED is connected inverted, it will light at 0V and will be off at 3.3v

therealprof (approved):
🎉

bors r+

Issue #406, opened 2022-02-22
serial::{Tx, Rx}::{listen, unlisten} are unsafe


AndreySmirnov81:
And also `Rx::{listen_idle, unlisten_idle}` are unsafe.

They all use `unsafe { (*USART::ptr()).cr1.modify(|_, w| w. ...) };`.

It is non-atomic read-modify-write.

Tx and Rx can be used simultaneously from different execution threads.
At the same time, they do not provide thread-safe access.

burrbull:
technically we could rewrite those with bit-banding (`bb::set`)

AndreySmirnov81:
I think this will be the most optimal solution.

PR #409, opened 2022-02-23
Serial rethinking

AndreySmirnov81:
1. Removed duplication of `Tx` and `Rx` functions from `Serial`.
Simplifying a `Serial` type function to store three elements: `Tx`, `Rx`, `ReleaseToken`.
(breaking change:
`serial.some_tx_fn` -> `serial.tx.some_tx_fn`
`serial.some_rx_fn` -> `serial.rx.some_rx_fn`).

2. Added the ability to release resources (USART, TX_PIN, RX_PIN).

3. `Tx` simultaneously implements `Write<u8>` and `Write<u16>`.
`Rx` simultaneously implements `Read<u8>` and `Read<u16>`.
Deleted the `with_u8_data` and `with_u16_data` functions.
(the `WORD` size must be derived from the context or specified explicitly (breaking change))

(some parts of the code are not deleted but commented out
in order to make it easier to view the changes,
they will be deleted in the next commit)

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -64,8 +64,6 @@
 //! let received: u8 = block!(rx.read()).unwrap();
 //!  ```
 
-#![deny(unsafe_op_in_unsafe_fn)]
-
 use core::convert::Infallible;
 use core::marker::PhantomData;
 use core::ops::Deref;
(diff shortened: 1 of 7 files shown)


src/serial.rs
```diff
@@ -411,77 +448,15 @@ where
...
     }
 }
 
 macro_rules! hal {
     ($USARTX:ident) => {
         impl Instance for $USARTX {
-            fn ptr() -> *const uart_base::RegisterBlock {
-                <$USARTX>::ptr() as *const _
+            unsafe fn steal() -> Self {
+                unsafe { crate::pac::Peripherals::steal().$USARTX }
```
burrbull, line 459: Using of `steal` more then once is UB.

AndreySmirnov81: Why using `steel` more than once is UB?
Isn't the use of constructions `*(<$USARTX>::ptr() as *const _)` and `crate::pac::Peripherals::steal().$USARTX` equivalent?

AndreySmirnov81: Note: the solution to issue #406 was not implemented in these changes in order not to complicate the understanding of the changes made. This will need to be solved with a separate commit.

AndreySmirnov81: We own a `USART` in two parts Tx and Rx (the issue with multithreaded access will be resolved later), so we do not store access to the `USART`, but get it as needed.

AndreySmirnov81: The idea to use `steal` arose when it was necessary to return `USART` from the `release` function.

AndreySmirnov81: I made the transition to using `steel` before I added `PinInfo`. Now it is possible to implement `USART` storage in `PinInfo` and rename `PinInfo` to `DataForRelease` or `ReleaseInfo`.

PR #414, opened 2022-03-06
Serial changes

AndreySmirnov81:
- `Serial::usart1/2/3` -> `Serial::new`
- Passing the `Clocks` parameter to `Serial` by reference
- `Serial` implements `Write<WORD>` and `Read<WORD>` for `WORD` simultaneously as u8 and u16
- Allow access to the `Tx` and `Rx` parts of the `Serial` without the need for splitting
- Allow `Serial` reconfiguration by reference to the `Tx` and `Rx` parts
- Allow `Serial` release after splitting
- Other various `Serial` improvements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Passing the `Clocks` parameter to `Serial` by reference
 - `Serial` implements `Write<WORD>` and `Read<WORD>` for `WORD` simultaneously as u8 and u16
   (the `WORD` must be derived from the context or specified explicitly)
+- Removed duplication of `Tx` and `Rx` functions from `Serial`.
+  serial.some_tx_fn -> serial.tx.some_tx_fn
+  serial.some_rx_fn -> serial.rx.some_rx_fn
 
 ## [v0.9.0] - 2022-03-02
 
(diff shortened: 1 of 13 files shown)


burrbull:
> * Removed duplication of `Tx` and `Rx` functions from `Serial`.
>       serial.some_tx_fn -> serial.tx.some_tx_fn
>       serial.some_rx_fn -> serial.rx.some_rx_fn

Return them back

and squash commits

AndreySmirnov81:
> and squash commits

I suggest leaving all commits. They allow you to understand what has changed and how.
Especially because of the commit "Formatting the `Serial` code (only changing in order)".

burrbull:
Have you tried to store `usart` instance inside `Tx` instead of just phantom?
I mean:
```rust
/// Serial transmitter
pub struct Tx<USART> {
    usart: USART,
}

/// Serial receiver
pub struct Rx<USART> {
    _usart: PhantomData<USART>,
}
```

AndreySmirnov81:
I cannot give preference to any of the Tx or Rx since they are equal in this respect, so let it be done, symmetrically.

burrbull:
Ok. LGTM.
If you need commit history, make branch copy.
Waiting for squashing.

AndreySmirnov81:
Are you sure you don't need a commits history?

AndreySmirnov81:
I want to make a squash for the last two commits after you look at the last commit.

PR #422, opened 2022-04-25
add temp/vref controls

fuchsnj:
Previously when an `Adc` is initialized, the temp/vref sensor was disabled. During an ADC read, the sensor has to be turned on and off, which significantly increases the time it takes for a read. Additionally, since there was no way to enable to sensor manually, I don't think it was possible to read VREF in DMA mode.

Now, there are 3 additional functions, `enable_temp_vref`, `disable_temp_vref` and `is_temp_vref_enabled`. These functions are optional, and will speed up reading from the ADC if it is enabled manually. 

The `read_aux` function now also only enables the sensor if the temp or vref channel is being read.

I tested this with an STM32F103 (bluepill)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - `Serial` support for UART4/5
 - Allow to set HSE bypass bit in `RCC` clock configuration register to use an external clock input on the `OSC_IN` pin [#485]
 - initial support of `embedded-hal-1.0` [#416]
+- Added ADC methods for temp/vref sensor control. The time required for reading from the ADC can be reduced using these [#422]
 - Add tools/check.py python script for local check [#467]
 - Add changelog check on PRs [#467]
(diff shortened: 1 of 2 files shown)


TheZoq2:
Thanks for the PR, and sorry for the review delay! Looks good to me, I left a small note on the return value of `enable_temp_vref` for discussion.

Could you also add this change to CHANGELOG.md?

fuchsnj:
I'm not seeing the note on the `enable_temp_vref` function, but I don't have strong opinions on how the return value is decided.

The options:
1. (current approach), return true if the function changed something, meaning the sensor was previously off.
2. Returns if the sensor was previously on.

I'm fine switching the approach if that sounds better, the one I picked was mostly arbitrary.

TheZoq2:
My question was more about if the function should return anything at all. Is there a use case for knowing if it was previously on (or off)?

fuchsnj:
> My question was more about if the function should return anything at all. Is there a use case for knowing if it was previously on (or off)?

Possibly. If that's blocking this change I'm fine removing the return type though.

PR #426, opened 2022-07-31
Bump bxcan

Fomys:
This pull request bump the version of bxcan from 0.6 to [0.7](https://github.com/stm32-rs/bxcan/releases/tag/v0.7.0), which introduce the second can FIFO. 
All CAN example are updated to use the new version of bxcan.
(I don't know how to add label, but this is a breaking change)


Fomys:
I don't know if this PR solve the issue https://github.com/stm32-rs/stm32f1xx-hal/issues/325, but I don't think it will be easy to support 0.6 and 0.7 at same time as the API has changed.

burrbull (approved):

PR #449, opened 2022-10-07
Bugfix: `CounterHz` and `Counter` `wait`

AndreySmirnov81:
`CounterHz` and `Counter` do not `wait` properly after the second and subsequent `start`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Allow `Serial` release after splitting.
 - `Spi::is_busy()`
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


burrbull:
It is strange. `trigger_update` should not cause interrupt.

https://github.com/stm32-rs/stm32f1xx-hal/blob/cafbac16d79da5c56bedb99b8c3f45b2c2463c38/src/timer.rs#L374-L379

AndreySmirnov81:
`wait` clears the flag correctly, however, the `timer` does not stop and works cyclically, which leads to the fact that the TIMx_SR.UIF flag is constantly set.

burrbull (approved):

PR #460, opened 2023-03-03
Allow GPIO splitting without reset and pin output configuration keeping initial state

### LINKED ISSUE #459: Read and keep initial GPIO output state
I would like to keep the initial state of a GPIO push-pull pin and read it from my program.

Currently the GPIO state is reset during the call to `split` and there is no method of the form `into_push_pull_output_with_current_state`. The later can be worked around by reading the output state directly from the ODR register and then calling `into_push_pull_output_with_state`.  However, I don't see a possibility to avoid the GPIO state reset with the current API during the call to `split`.

surban:
Fixes #459.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -119,8 +119,13 @@ pub trait GpioExt {
     /// The to split the GPIO into
     type Parts;
 
-    /// Splits the GPIO block into independent pins and registers
+    /// Splits the GPIO block into independent pins and registers.
+    ///
+    /// This resets the state of the GPIO block.
     fn split(self) -> Self::Parts;
+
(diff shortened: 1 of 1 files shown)


src/gpio.rs
```diff
@@ -391,6 +396,19 @@ macro_rules! gpio {
                         )+
                     }
                 }
+
+                fn split_without_reset(self) -> Parts {
```
burrbull, line 400:
```suggestion
                unsafe fn split_without_reset(self) -> Parts {
```

surban: Fixed

burrbull (approved):

PR #462, opened 2023-04-12
remap enums

burrbull:
This is breaking change.
It uses similar idea to https://github.com/stm32-rs/stm32f4xx-hal/pull/594 and family, but different.
Istead of `pins: TupleOfPinEnums` peripheral keeps `pins: EnumOfPinTupleCombinations`. It is less beautiful and flexible, but gives workaround for `AFIO` map issues (which is F1 only).

Needs complex testing. cc @all_who_can_test_and_have_ideas :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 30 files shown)


therealprof (approved):
LGTM, not in a position to test this ATM, though.

PR #468, opened 2023-08-01
Fix RTC selection of frequency for LSI and HSE

Rutherther:
So far, `Rtc::select_frequency` did not consider frequencies other than LSE_HERTZ, which was valid only for LSE with 32.768 kHz. For LSI, which has 40 kHz, and HSE, which may have an arbitrary frequency in a given range, the frequency was selected based on the wrong RTC frequency.

This PR aims to fix that. I've also refactored the new methods a bit, using the select_frequency instead of repeating the same code.

I've solved this by adding frequency field to the Rtc, there is no other way I could think of, as HSE frequency may be different depending on the crystal selected for specific application. As far as I understand it, LSE does not necessarily have to be 32.768 kHz, so maybe adding the possibility to change the frequency of LSE could prove useful as well? Adding the frequency field unfortunately introduces redundancy for LSI and (maybe) LSE.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Update `bxcan`, `heapless`, `mfrc522`, reenable `mpu9250` example [#513]
 - PWM timer auto reload value is now preloaded/buffered [#453]
 - Move from bors/manual merge to GH merge queue [#467]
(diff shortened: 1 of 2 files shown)


Rutherther:
In my previous PR #466, I forgot to add the change to CHANGELOG.md, this PR adds that change to CHANGELOG as well.

tmehr-enqt (commented):
Looks good to me, didn't test it but the logic looks fine.

burrbull (approved):

PR #479, opened 2024-01-25
Reexport Direction from older embedded_hal

samueltardieu:
`embedded_hal::Direction` has been removed from recent `embedded_hal` releases. It needs to be reexported from `qei` in order to allow users of `Qei::direction()` to be able to match against the method result if they use a recent `embdded_hal` release.


samueltardieu:
Note that all the CI failures are unrelated to this patch and also happen on master.

samueltardieu:
Rebased on current master (which now compiles).

burrbull (approved):

PR #480, opened 2024-01-30
Replace UB code by a legitimate pointer access

samueltardieu:
This removes a compilation error due to the `invalid-reference-casting` check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Move from bors/manual merge to GH merge queue
 - Add tools/check.py python script for local check
 - Add changelog check on PRs
+- Replace UB code by a legitimate pointer access
 
 ## [v0.10.0] - 2022-12-12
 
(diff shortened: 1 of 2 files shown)


adamgreig:
In this case the error is legit and the code has UB. It should be updated to at least use `&mut self.spi.dr as *mut FrameSize` to avoid casting a shared reference to a mutable pointer. Better yet is to use `addr_of!()` as was done here: https://github.com/stm32-rs/stm32f0xx-hal/pull/170

samueltardieu:
> In this case the error is legit and the code has UB. It should be updated to at least use `&mut self.spi.dr as *mut FrameSize` to avoid casting a shared reference to a mutable pointer. Better yet is to use `addr_of!()` as was done here: [stm32-rs/stm32f0xx-hal#170](https://github.com/stm32-rs/stm32f0xx-hal/pull/170)

`addr_of!()` is the way to go, `&mut` is not possible here as `DerefMut` is not implemented.

burrbull (approved):

PR #483, opened 2024-03-21
Add dac

apeng2012:
clone from [stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal)


burrbull:
Fix "unnecessary `unsafe`" warning, please.

Issue #481, opened 2024-02-06
Add possibility to set HSE bypass bit (HSEBYP) in RCC_CR register


chrisegerer:
On the STM32F1 series, the HSE clock can be generated either from external crystal or an external clock. If an external clock source is used, the HSEBYP bit must be set in the RCC_CR clock control register (see chapter 7.2.1 in STM32F1 reference manual). This is a common setup on ST's Nucleo dev boards, where the external clock is provided by the integrated ST-LINK debugger probe.

Currently, the stm32f1xx-hal does not allow to set the HSEBYP bit. This limits the clock setup to only use the HSI clock. Other STM32 HALs implement setting this bit.

chrisegerer:
Implemented in this [PR](https://github.com/stm32-rs/stm32f1xx-hal/pull/485)

PR #485, opened 2024-04-15
Allow to set HSE bypass bit in clock config

chrisegerer:
Allows to set HSE bypass bit in RCC clock config register to use external clock supplied through OSC_IN pin.
Implementation is adopted from stm32f4xx-hal.
Related to #481


chrisegerer:
@therealprof: please assign to someone for review

burrbull (approved):

PR #510, opened 2024-10-08
Use latest version of `usb-device` and `usbd-serial`

eivindbergem:
Use the latest version to get implementations of `Read` and `Write` from `embedded-io` for `SerialPort`.


burrbull:
CI fails. See https://github.com/stm32-rs/stm32f4xx-hal/pull/699/files for how to solve.

burrbull (approved):

PR #517, opened 2024-10-16
init channels after Pwm

burrbull:
See https://github.com/stm32-rs/stm32f4xx-hal/pull/791

~~Depends on #553~~


burrbull:
Rebased. Should be easier to review.

burrbull:
@tmehr-enqt Review this one too, please.

tmehr-enqt (approved):

PR #520, opened 2024-12-14
I2c::new remap

burrbull:


burrbull:
@pdgilbert Test this yet one time.

pdgilbert:
Ok. I updated to new `rmp-new` commit:
```
Updating stm32f1xx-hal v0.10.0 (https://github.com/stm32-rs/stm32f1xx-hal?branch=rmp-new#04b84469) -> #d9f7eab0
```
and re-ran my example test with success. I also ran a much larger set of example tests with 60 of 80 passing. That is pretty good, since I am still cleaning up many of the tests in the switch to eh-1.  (It is up from 5 of 80 a few days ago.) The CI for these is at https://github.com/pdgilbert/rust-integration-testing/actions. The overall coverage of my examples may not be very good, but I think it is pretty good for `I2C`. (So I give an approving review.)  I will do some hardware run testing soon.

pdgilbert (approved):
I've tested I2C changes on several examples and they work well.

PR #526, opened 2025-01-21
Add enable/disable EOC interrupt functions for ADCs

Tnze:
This patch allow user enable End of Convert interrupt for ADCs by calling `adc1.enable_eoc_interrupt()`.

Tested on STM32F103CBT6, after enabling the `eocie` register. The interrupt `ADC1_2` is triggered every time ADC convert is finished. This allows more efficient waiting for ADC result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Add an option to allow overclocking [#494]
 - `new` on gpio mode [#506]
 - Add `Serial` `rx`/`tx` constructors [#509]
+- Add enable/disable EOC interrupt functions for ADCs [#526]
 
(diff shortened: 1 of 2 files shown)


burrbull (approved):

Tnze:
@burrbull To make this useful. Addition changes are required.

We need a way taking apart `convert()` logic, and this is my temporary solution:
https://github.com/Tnze/stm32f1xx-hal/commit/6da9c4f429433a9d557263ccf76c7d60c38811a7

There is obvious problem of this and it will be good to have someone to design a better API.

Tnze:
I just realized that `read()` function returns `nb::Result` which the API needed here. Make it returns [WouldBlock](https://docs.rs/nb/latest/nb/enum.Error.html) when the ADC convert not finished. But it will changes the behavior and all previous users need to wrap `read()` calls with `nb::block!()` to match the previous behavior.

PR #542, opened 2025-08-29
SPI `FrameSize` and missing e-hal 1.0

burrbull:
This is fast impl. Not tested.

cc @everdrone


tmehr-enqt:
What tests do you think have to be done? I did some quick tests with an spi loopback on a f103 nucleo. That worked. If you assume that the peripheries of the F series are basically all the same I would say it works.

tmehr-enqt:
I have now. Still as expected.

tmehr-enqt:
Maybe also add a allow for static_mut_refs so the spi-slave example doesn't generate warnings?

PR #543, opened 2025-08-30
backport I2c implementation from f4xx-hal

burrbull:
Depends on #542


tmehr-enqt:
In my opinion this is ready. It's also the last todo grep found.

PR #545, opened 2025-09-05
I2C: fixed example and added reset for blocking

### LINKED ISSUE #544: exemple i2c-bme280 broken, no reset on blocking I2C
I noticed the method I2C::blocking_i2c is gone, this breaks the i2c-bme280 example. I think reintroducing that constructor would be good.

Also the now public method I2C::reset has no equivalent for BlockingI2c.

tmehr-enqt:
This resolves #544.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


burrbull:
> @burrbull I would like that. Work uses this, so I think I will get the time to do maintenance and stuff.

cc @therealprof @adamgreig

PR #554, opened 2025-11-30
DMA demacrofy

burrbull:


tmehr-enqt (approved):
Looks good. Less macros is always a good idea, especially if the code gets shorter that way. There seems to be a conflict in spi.rs

PR #565, opened 2026-05-08
Support Latest Embedded-io

ScottGibb:
Thanks for making this cool HAL, I would like to contribute this small PR to bump the embedded-io dependency so that we can link it with the latest drivers using that trait. For example im building this [Power Monitor driver](https://github.com/ScottGibb/JSY-MK-194-rs) at the moment and would like to test it on an STM32 Blue Pill


burrbull (approved):

burrbull:
Describe changes in Changelog.md, please.

burrbull (approved):

